PR #18, opened 2016-02-02
test ReadMatrixFromEnviron

HelloGrayson:
cc @yarpc/golang

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,12 +5,12 @@ project := xlang
 install:
 	glide --version || go get github.com/Masterminds/glide
 	GO15VENDOREXPERIMENT=1 glide install
-	GO15VENDOREXPERIMENT=1 go build `GO15VENDOREXPERIMENT=1 glide novendor`
+	GO15VENDOREXPERIMENT=1 go build `glide novendor`
 
 
 .PHONY: test
 test:
-	GO15VENDOREXPERIMENT=1 go test `GO15VENDOREXPERIMENT=1 glide novendor`
+	GO15VENDOREXPERIMENT=1 go test `glide novendor`
 
 
 .PHONY: xlang
--- a/environ_test.go
+++ b/environ_test.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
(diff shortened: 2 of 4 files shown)


environ_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadMatrixFromEnviron(t *testing.T) {
+
+	clients := []string{"yarpc-go", "yarpc-node", "yarpc-browser"}
+	servers := []string{"yarpc-go", "yarpc-node"}
+	transports := []string{"http", "tchannel"}
+
+	os.Setenv("XLANG_CLIENTS", strings.Join(clients, ","))
```
HelloGrayson: I might switch this to `os.Setenv("XLANG_CLIENTS", "yarpc-go,yarpc-node,yarpc-browser")` to favor readability, and just duplicate the values in the assert statement.

blampe:
:sheep:

HelloGrayson:
I do not know what this meenz :P

Makefile
```diff
@@ -5,12 +5,12 @@ project := xlang
 install:
 	glide --version || go get github.com/Masterminds/glide
 	GO15VENDOREXPERIMENT=1 glide install
-	GO15VENDOREXPERIMENT=1 go build `GO15VENDOREXPERIMENT=1 glide novendor`
+	GO15VENDOREXPERIMENT=1 go build `glide novendor`
```
prashantv, line 8: you should just set this at the top of the Makefile, and then all commands should inherit it.

HelloGrayson, line 8: GNU Make allows for top-level env var's? I didn't realize that... or did you just mean something like:

```
env:
    export GO15VENDOREXPERIMENT=1

install: env
    ...
```

prashantv, line 8: You can do top level exports,
https://github.com/uber/tchannel-go/blob/dev/Makefile#L12

Each command for a target runs in its own sub-shell so having an `env` target won't work (it's lost after that shell). The export command is what you want to use.

Then you don't need to keep setting the variable everywhere.

PR #19, opened 2016-02-02
WIP - lint during build time

HelloGrayson:
cc @yarpc/golang


HelloGrayson:
@akshayjshah @prashantv what's the "right" go way to run `golint` and `go vet` from CI/Travis?

Install directly from .travis.yml with `go get` or vendor the deps?

PR #20, opened 2016-02-03
set GOVENDOR15EXPERIMENT once in Makefile

HelloGrayson:
cc @yarpc/golang


prashantv:
title / commit message have a typo, should be "GOVENDOR15EXPERIMENT". otherwise lgtm.

HelloGrayson:
@prashantv good catch, fixed.

PR #24, opened 2016-02-09
fix scratch image

HelloGrayson:
adding the examples dir resulted in the scratch image failing:

```
$ CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o main `glide novendor`
go build: cannot use -o with multiple packages
```

cc @kriskowal


kriskowal:
ty

PR #36, opened 2016-03-16
Always output crossdock logs during Travis builds

### LINKED ISSUE #33: Run logs if build fails
We should output the logs if a build fails, this will help understand what went wrong.

see: https://github.com/docker/compose/issues/3058#issuecomment-191031502

HelloGrayson:
Fixes #33


kriskowal:
:ok:

PR #38, opened 2016-03-16
Call deadlines

HelloGrayson:
Instead of hanging forever when a test client hangs, give them 5 seconds to respond.

cc @yarpc/yarpc @abhinav @prashantv

### DIFF
--- a/execute/run.go
+++ b/execute/run.go
@@ -7,10 +7,10 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/yarpc/crossdock/plan"
+
 	"golang.org/x/net/context"
 	"golang.org/x/net/context/ctxhttp"
-
-	"github.com/yarpc/crossdock/plan"
 )
 
 // Run the test program for a given Plan
--- a/execute/run.go
+++ b/execute/run.go
@@ -4,14 +4,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"net/url"
+	"time"
+
+	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
 
 	"github.com/yarpc/crossdock/plan"
 )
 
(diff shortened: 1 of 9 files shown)


plan/config.go
```diff
@@ -2,19 +2,26 @@ package plan
...
-	const axisKeyPrefix = "CROSSDOCK_AXIS_"
-
-	clients := strings.Split(os.Getenv(clientsKey), ",")
-	clients = trimCollection(clients)
+const callDeadlineDefault = 5
 
+// ReadConfigFromEnviron creates a Config by looking for CROSSDOCK_ environment vars
+func ReadConfigFromEnviron() *Config {
+	const (
+		callDeadlineKey = "CROSSDOCK_CALL_DEADLINE"
```
kriskowal: Might be less confusing to call it TTL, since it’s a duration relative to start time. Deadline implies an absolute time.

kriskowal:
:ok:

jc-fireball:
it is one client is unreachable then all fail, or let the crossdock test continue with it.

HelloGrayson:
@junchaowu just that one call will timeout, and the test will continue.

plan/config.go
HelloGrayson: Renaming to `CROSSDOCK_CALL_TIMEOUT`.

execute/run.go
```diff
@@ -4,14 +4,17 @@ import (
...
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"net/url"
+	"time"
+
+	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
 
 	"github.com/yarpc/crossdock/plan"
```
abhinav: Nit: Fix import group ordering. It should be:

   std library

```
local packages

third party packages
```

abhinav:
:+1:

Issue #51, opened 2016-05-12
Minimal Reporter


HelloGrayson:
Let's implement a new reporterer `minimal` - see [`Reporter interface` here](https://github.com/yarpc/crossdock/blob/master/output/report.go#L31).

```
Starting Crossdock...

RAW .....
JSON .............
THRIFT .........................................
ERRORS .......................................................
HEADERS ...........F..........SS...........

100 passed
1 failed
2 skipped

F - {0xc820012f60 java map[server:go transport:tchannel behavior:raw]} - <output>
```

Note the errors listed in verbose line in the bottom.

abhinav:
👍  except `s/DEBUG/VERBOSE/` and expanding on the exact parameters passed to a client in a deterministic order. Something like,

```
F - behavior=raw client=java server=go transport=tchannel: <output>
```

yurishkuro:
+1 on sorting the parameter names. I thought I saw a table format once, or was it just in the readme?

HelloGrayson:
#52 will take care of the counts

HelloGrayson:
@yurishkuro the table format was tossed because it was incompatible with streaming the results - we'll add it back later

PR #53, opened 2016-05-13
Allow multiple console reporters

HelloGrayson:
Setting `REPORT` to a valid reporter value will result in a different console reporter.

The default is `list`, and `REPORT` is optional.

This paves the way for #51 

@yarpc/golang

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,7 +46,7 @@ func main() {
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	reporter := output.GetReporter("list")
+	reporter := output.GetReporter(config.Report)
 	summary := reporter.Stream(results)
 	output.Summarize(summary)
 
--- a/main.go
+++ b/main.go
@@ -46,7 +46,11 @@ func main() {
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	reporter := output.GetReporter(config.Report)
+	reporter, err := output.GetReporter(config.Report)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	summary := reporter.Stream(results)
 	output.Summarize(summary)
 
--- a/output/list.go
+++ b/output/list.go
@@ -18,39 +18,40 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package main
+package output
 
 import (
 	"fmt"
-	"log"
-	"os"
-	"time"
 
 	"github.com/yarpc/crossdock/execute"
-	"github.com/yarpc/crossdock/output"
-	"github.com/yarpc/crossdock/plan"
+
+	"github.com/fatih/color"
 )
 
-func main() {
-	fmt.Printf("\nCrossdock starting...\n\n")
+var green = color.New(color.FgGreen).SprintFunc()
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -46,7 +46,12 @@ func main() {
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	summary := output.Stream(results)
+	reporter, err := output.GetReporter(config.Report)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
akshayjshah, line 52: Assuming that the output has to go to stdout? If not, `log.Fatalln` will do this for you.

output/report.go
```diff
@@ -0,0 +1,55 @@
...
+func (f ReporterFunc) Stream(tests <-chan execute.TestResponse) Summary {
+	return f(tests)
+}
+
+// GetReporter returns a ReporterFunc for the given name
+func GetReporter(name string) (Reporter, error) {
+	// default reporter
+	if name == "" {
+		return ReporterFunc(List), nil
+	}
```
akshayjshah, line 47: Why not put this in the switch statement below?

output/summary.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "fmt"
+
+// Summary contains an account of the test run
+type Summary struct {
+	Failed bool
+
+	SuccessAmount int
+	FailAmount    int
+	SkippedAmount int
```
akshayjshah: nittiest nit ever: "amount" is for uncountable things, "number" is for countable things (e.g., "amount of sugar", "number of teaspoons of sugar"). At some point, might be nice to rename to `NumSuccessful`, etc..

main.go
HelloGrayson, line 52: I don't want the timestamp prefix'd for now, so I'm being lazy and doing it this way till I configure a logger :)

output/report.go
HelloGrayson, line 47: Foreshadowing - once there's 2 reporters, which will happen in #51, then we'll want people to be able to select one.

output/summary.go
HelloGrayson: I've never heard that before - thanks for the heads up 👍

output/list.go
```diff
@@ -32,17 +32,8 @@ var green = color.New(color.FgGreen).SprintFunc()
...
-
-	SuccessAmount int
-	FailAmount    int
-	SkippedAmount int
-}
-
-// Stream results to the console, error at end if any fail
-func Stream(tests <-chan execute.TestResponse) Summary {
+// List is a ReporterFunc that produces a verbose list of results
+func List(tests <-chan execute.TestResponse) Summary {
```
abhinav: FYI: List is not actually a ReporterFunc. You can optionally do the following to actually make it a `ReporterFunc` so that you don't have to cast it later.

``` go
var List ReporterFunc = func(tests <-chan executer.TestResponse) Summary {
  // ...
}
```

output/report.go
```diff
@@ -0,0 +1,55 @@
...
+	// else a specific value was provided
+	switch name {
```
abhinav, line 49: probably do a `strings.ToLower(name)` first so that you can do `REPORT=LIST`

abhinav:
LGTM

output/report.go
HelloGrayson, line 49: Going to `ToLower` when the config is created so the whole program can rely on the format

HelloGrayson:
@akshayjshah @abhinav feedback addressed, thx sirs 😄

PR #54, opened 2016-05-20
Add a JSON report feature

kriskowal:
r @breerly @yarpc/golang

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,10 @@ crossdock:
       - BEHAVIOR_DANCE=client,server,transport
       - BEHAVIOR_SING=client,server,transport
 
+      - REPORT=json
+    volumes:
+      - .crossdock:/crossdock
+
 alpha:
     image: breerly/hello-server
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,8 @@ crossdock:
       - BEHAVIOR_SING=client,server,transport
 
       - REPORT=json
+      - JSON_REPORT_PATH=/crossdock/report.json
+
     volumes:
       - .crossdock:/crossdock
 
--- a/execute/entities.go
+++ b/execute/entities.go
@@ -20,7 +20,10 @@
 
 package execute
 
-import "github.com/yarpc/crossdock/plan"
+import (
+	"fmt"
+	"github.com/yarpc/crossdock/plan"
+)
 
 // TestResponse contains the reply from a test client, most importantly,
 // contains a list of Results for the test cases ran by the test client
@@ -48,3 +51,17 @@ const (
 	// Skipped indicates a client' TestCase did not run
 	Skipped
 )
+
+var StatusNames = map[Status]string{
+	Success: "success",
+	Failed:  "failed",
+	Skipped: "skipped",
+}
+
+func (s Status) MarshalText() (text []byte, err error) {
+	name, ok := StatusNames[s]
+	if !ok {
+		return nil, fmt.Errorf("unexpected status %d", s)
+	}
+	return []byte(name), nil
+}
--- a/execute/entities.go
+++ b/execute/entities.go
@@ -22,6 +22,7 @@ package execute
 
 import (
 	"fmt"
+
 	"github.com/yarpc/crossdock/plan"
 )
 
@@ -52,16 +53,19 @@ const (
 	Skipped
 )
 
-var StatusNames = map[Status]string{
-	Success: "success",
-	Failed:  "failed",
-	Skipped: "skipped",
+func (s Status) String() string {
+	switch s {
+	case Success:
+		return "success"
+	case Failed:
+		return "failed"
+	case Skipped:
+		return "skipped"
+	default:
+		return fmt.Sprintf("Status(%v)", s)
+	}
 }
 
-func (s Status) MarshalText() (text []byte, err error) {
-	name, ok := StatusNames[s]
-	if !ok {
-		return nil, fmt.Errorf("unexpected status %d", s)
-	}
-	return []byte(name), nil
+func (s Status) MarshalText() ([]byte, error) {
+	return []byte(s.String()), nil
 }
--- a/output/json.go
+++ b/output/json.go
@@ -7,16 +7,10 @@ import (
 	"io/ioutil"
 )
 
-var StatusNames = map[execute.Status]string{
-	execute.Success: "success",
-	execute.Failed:  "failed",
-	execute.Skipped: "skipped",
-}
-
 type JSONTestReport struct {
 	Client    string         `json:"client"`
 	Arguments plan.Arguments `json:"arguments"`
-	Status    string         `json:"status"`
+	Status    execute.Status `json:"status"`
 	Output    string         `json:"output"`
 }
 
@@ -55,7 +49,7 @@ var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestRespo
 			behaviorReport.Tests = append(behaviorReport.Tests, JSONTestReport{
 				Client:    client,
 				Arguments: args,
-				Status:    StatusNames[result.Status],
+				Status:    result.Status,
 				Output:    result.Output,
 			})
 		}
--- a/output/json.go
+++ b/output/json.go
@@ -2,9 +2,10 @@ package output
 
 import (
 	"encoding/json"
+	"io/ioutil"
+
 	"github.com/yarpc/crossdock/execute"
 	"github.com/yarpc/crossdock/plan"
-	"io/ioutil"
 )
 
 type JSONTestReport struct {
--- a/output/json.go
+++ b/output/json.go
@@ -26,8 +26,9 @@ type JSONReport struct {
 
 var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
 	summary := Summary{}
-	report := JSONReport{}
-	report.Behaviors = make(map[string]*JSONBehaviorReport)
+	report := JSONReport{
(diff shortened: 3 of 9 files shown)


output/json.go
```diff
@@ -0,0 +1,74 @@
...
+	"io/ioutil"
+)
+
+var StatusNames = map[execute.Status]string{
+	execute.Success: "success",
+	execute.Failed:  "failed",
+	execute.Skipped: "skipped",
+}
+
+type JsonTestReport struct {
```
prashantv: nit: here and below, go prefers to use `JSON` or `json`,
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

(That whole page is a good read)

output/json.go
```diff
@@ -0,0 +1,74 @@
+package output
+
+import (
+	"encoding/json"
+	"github.com/yarpc/crossdock/execute"
+	"github.com/yarpc/crossdock/plan"
+	"io/ioutil"
+)
+
+var StatusNames = map[execute.Status]string{
```
prashantv: nit: can we implement a `String` method on the `execute.Status` type instead?

If we also implement `MarshalText`, then the same type can be used in the `JSONTestReport`

output/json.go
```diff
@@ -0,0 +1,74 @@
...
+	Behaviors map[string]*JsonBehaviorReport `json:"behaviors"`
+}
+
+var Json ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) (summary Summary) {
+
+	report := JsonReport{}
+	report.Behaviors = make(map[string]*JsonBehaviorReport)
+
+	for _, behavior := range config.Behaviors {
+		behaviorReport := new(JsonBehaviorReport)
```
prashantv: nit: you can do

``` go
behaviorReport := &JSONBehaviorReport{
  Tests: [...],
  Params: behavior.Params,
}
```

output/json.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	err = ioutil.WriteFile("/crossdock/crossdock.json", data, 0644)
+	if err != nil {
+		panic(err)
+	}
+
+	return
```
prashantv: nit: avoid naked named + returns
https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

output/json.go
HelloGrayson: Lint should catch this... perhaps a good opportunity to make sure we're running that with tests...

output/json.go
HelloGrayson: +1 - if this isn't an early return then probably not necessary

docker-compose.yml
```diff
@@ -13,6 +13,10 @@ crossdock:
       - BEHAVIOR_DANCE=client,server,transport
       - BEHAVIOR_SING=client,server,transport
 
+      - REPORT=json
```
HelloGrayson, line 16: This is really cool! 

I think what we'll want to do is support multiple reporters, so that you don't lose the console output:

```
- REPORT=list,json
```

output/json.go
```diff
@@ -0,0 +1,74 @@
...
+			})
+		}
+	}
+
+	data, err := json.Marshal(report)
+	if err != nil {
+		panic(err)
+	}
+
+	err = ioutil.WriteFile("/crossdock/crossdock.json", data, 0644)
```
HelloGrayson: This should come from a config var.

docker-compose.yml
HelloGrayson, line 16: Created #55

output/json.go
kriskowal: Yeah, saw that. Will change.

Tell me more about the proper style for, say, “XML HTTP request”.

output/json.go
kriskowal: The explicit return is necessary for named return variables. However, named return variables aren’t necessary. :)

output/json.go
```diff
@@ -0,0 +1,75 @@
+package output
+
+import (
+	"encoding/json"
+	"github.com/yarpc/crossdock/execute"
+	"github.com/yarpc/crossdock/plan"
+	"io/ioutil"
```
abhinav: [Import group ordering](https://github.com/golang/go/wiki/CodeReviewComments#imports)

You want,

``` go
import (
  // standard library

  // local packages

  // third party packages
)
```

`goimports` will do part of this for you if you have it set up.

output/json.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var StatusNames = map[execute.Status]string{
+	execute.Success: "success",
+	execute.Failed:  "failed",
+	execute.Skipped: "skipped",
+}
+
+type JSONTestReport struct {
```
abhinav: Generally, you want to avoid stuttering in exported types. Packages that import the `json` package will use this as `json.JSONTestReport`. It is completely acceptable to call it `TestReport` instead so that usage is `json.TestReport`. Same for `JSONBehaviorReport` and `JSONReport`.

abhinav: Although, if the usage of this package doesn't involve actually accessing these types, you should make them private (`testReport` or `jsonTestReport`).

output/json.go
prashantv: It's an angry identifier, `XMLHTTPRequest`. However, since there's packages, you would use that to namespace your identifiers into something like: `xml/http` and have a `http.Request`, or `xmlhttp.Request`

output/json.go
abhinav: Oh, my bad. Ignore these comments.

output/json.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type JSONReport struct {
+	Behaviors map[string]*JSONBehaviorReport `json:"behaviors"`
+}
+
+var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
+	summary := Summary{}
+	report := JSONReport{}
+	report.Behaviors = make(map[string]*JSONBehaviorReport)
```
abhinav: nit:

``` go
report := JSONReport{Behaviors: make(map[string]*JSONBehaviorReport)}
```

plan/config.go
```diff
@@ -57,12 +59,19 @@ func ReadConfigFromEnviron() (*Config, error) {
 			behaviors[behavior.Name] = behavior
 		}
 	}
+
+	jsonReportPath := os.Getenv(jsonReportPathKey)
+	if jsonReportPath == "" {
+		jsonReportPath = "/crossdock/report.json"
```
abhinav, line 65: Should we make the environment variable required instead? No JSON reporting if the report path isn't set? Sending something to `/crossdock` is undesirable for when you're running it directly outside of docker.

abhinav, line 65: (Whether we want silent downgrade to regular console reporting or loud failure is up to you.)

HelloGrayson:
I like the idea of making all reporters go to the console by default, but allowing a file to be specified as @prashantv mentioned:

```
- REPORT=list,json(file="/crossdock/report.json")
```

We should probably get this API settled before landing - lets talk in person :)

kriskowal:
This is ready for another pass and hopefully +1s. I will follow up with two PR’s, one to refactor reporters so they don’t necessarily involve channels, and another to support REPORT=list,json. I will probably do that quick and dirty and do a proper mux later.

plan/config.go
kriskowal, line 65: I’ll follow up with that change with the coming refactor to reporters so they don’t expose channels in their interface.

kriskowal:
@breerly I like the idea, though we can stay out of the parsing game by using moar environment variables. I would like to get us there incrementally.

kriskowal:
I intend to follow up with #56 and #57 to address remaining issues with small diffs.

execute/entities.go
```diff
@@ -48,3 +51,17 @@ const (
 	// Skipped indicates a client' TestCase did not run
 	Skipped
 )
+
+var StatusNames = map[Status]string{
+	Success: "success",
+	Failed:  "failed",
+	Skipped: "skipped",
+}
```
abhinav, line 67: I think what you want to do here instead is,

``` go
func (s Status) String() string {
  switch s {
    case Success: return "success"
    // ..
    default: return fmt.Sprintf("Status(%v)", int(s))
  }
}
```

and use that in `MarshalText()`

HelloGrayson, line 67: +1

abhinav:
LGTM besides new and existing nits

output/list.go
```diff
@@ -33,7 +34,7 @@ var yellow = color.New(color.FgYellow).SprintFunc()
 var red = color.New(color.FgRed).SprintFunc()
 
 // List is a ReporterFunc that produces a verbose list of results
-var List ReporterFunc = func(tests <-chan execute.TestResponse) Summary {
+var List ReporterFunc = func(_ *plan.Config, tests <-chan execute.TestResponse) Summary {
```
HelloGrayson: might as well name the config `config`, no reason to `_` here.

HelloGrayson:
fair enough - we can use multiple env vars for now.

🚢 🐑

execute/entities.go
kriskowal, line 67: Tried it. JSON marshal doesn’t pick it up. I’ll push a switch/case variant of MarhsalText and use that instead of the table, but the String overload doesn’t fly.

prashantv, line 67: You need to implement `String` and `MarshalText`, you should do something like:

``` go
func (s Status) String() string {
  switch s {
  case Success: return "success"
  [...]
  default: return fmt.Sprintf("Status(%v)",int(s))
  }
}


func (s Status) MarshalText() (text []byte, error) {
  return []byte(s.String()), nil
}
```

PR #56, opened 2016-05-23
Remove channels from the reporter interface

kriskowal:
This is to pave the way for multiplexing reporters without having to use concurrency in the mux layer.

Intend to rebase after #54 lands.

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,11 +52,19 @@ func main() {
 		os.Exit(1)
 	}
 
-	reporter.Start(config)
+	err = reporter.Start(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	for test := range results {
 		reporter.Next(test, config)
 	}
-	summary := reporter.End(config)
+	summary, err := reporter.End(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	output.Summarize(summary)
 
 	if summary.Failed == true {
--- a/main.go
+++ b/main.go
@@ -51,7 +51,12 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(config, results)
+
+	reporter.Start(config)
+	for test := range results {
+		reporter.Next(test, config)
+	}
+	summary := reporter.End(config)
 	output.Summarize(summary)
 
 	if summary.Failed == true {
--- a/output/json.go
+++ b/output/json.go
@@ -24,40 +24,26 @@ type JSONReport struct {
 	Behaviors map[string]*JSONBehaviorReport `json:"behaviors"`
 }
 
-var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
-	summary := Summary{}
-	report := JSONReport{
-		Behaviors: make(map[string]*JSONBehaviorReport),
-	}
+type JSON struct {
+	s Summary
+	r JSONReport
+}
+
+func (j *JSON) Start(config *plan.Config) {
+	j.r.Behaviors = make(map[string]*JSONBehaviorReport)
 
 	for _, behavior := range config.Behaviors {
 		behaviorReport := &JSONBehaviorReport{
 			Tests:  make([]JSONTestReport, 0, 10),
 			Params: behavior.Params,
 		}
-		report.Behaviors[behavior.Name] = behaviorReport
+		j.r.Behaviors[behavior.Name] = behaviorReport
 	}
 
-	for test := range tests {
-		client := test.TestCase.Client
-		args := test.TestCase.Arguments
-		behavior := test.TestCase.Arguments["behavior"]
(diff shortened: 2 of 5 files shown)


main.go
```diff
@@ -52,11 +52,19 @@ func main() {
 		os.Exit(1)
 	}
 
-	reporter.Start(config)
+	err = reporter.Start(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
abhinav: nit: can inline err

``` go
if err := reporter.Start(config); err != nil {
    // ...
}
```

main.go
```diff
@@ -52,11 +52,19 @@ func main() {
...
+	}
 	for test := range results {
 		reporter.Next(test, config)
 	}
-	summary := reporter.End(config)
+	summary, err := reporter.End(config)
+	if err != nil {
```
abhinav: Same as above: can inline err

main.go
```diff
@@ -51,7 +51,20 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(results)
+
+	err = reporter.Start(config)
+	if err != nil {
```
abhinav: Repeating comment here because GH is hiding my comment for some reason:

nit: can inline err

``` go
if err := reporter.Start(config); err != nil {
    // ...
}
```

abhinav:
Reviewed new changes added on top of #54. LGTM

kriskowal:
This diff is now rebased on master, @breerly.

HelloGrayson:
I think we should make the summary its own reporter - that way each reporter doesnt have to do the summary bookeeping:

```
- REPORT=list,summary
```

output/json.go
```diff
@@ -24,48 +25,58 @@ type JSONReport struct {
...
-		Behaviors: make(map[string]*JSONBehaviorReport),
+type JSON struct {
+	s Summary
+	r JSONReport
+}
+
+func (j *JSON) Start(config *plan.Config) error {
+	j.r.Behaviors = make(map[string]*JSONBehaviorReport)
+
+	if config.JSONReportPath == "" {
```
HelloGrayson, line 36: There's a default, no?

HelloGrayson, line 36: I suppose being defensive is best :)

HelloGrayson:
Could you explain why the `Start()`, `Next()` and `End()` interaction is better than calling `Reporter.Stream()`?

kriskowal:
@breerly The Stream() interface depends on using channels. Having a channel for every reporter, and a mux that asynchronously pumps results into each of those reporters is unnecessarily asynchronous. Using a synchronous generator interface instead of channels can hide the detail of channels if they become necessary.

kriskowal:
Added #58 to factor out the summary generator

output/report.go
```diff
@@ -29,29 +29,23 @@ import (
...
-	Stream(config *plan.Config, tests <-chan execute.TestResponse) Summary
-}
-
-// ReporterFunc streams test results to the console
-type ReporterFunc func(config *plan.Config, tests <-chan execute.TestResponse) Summary
-
-// Stream allows ReporterFunc to fulfill the Reporter interface
-func (f ReporterFunc) Stream(config *plan.Config, tests <-chan execute.TestResponse) Summary {
-	return f(config, tests)
+	Start(config *plan.Config) error
```
HelloGrayson, line 32: config doesn't change over the course of the reporter - shouldn't we just instantiate the reporter with a config instead of passing it into each of these methods?

output/json.go
```diff
@@ -24,48 +25,58 @@ type JSONReport struct {
 	Behaviors map[string]*JSONBehaviorReport `json:"behaviors"`
 }
 
-var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
-	summary := Summary{}
-	report := JSONReport{
-		Behaviors: make(map[string]*JSONBehaviorReport),
+type JSON struct {
+	s Summary
```
HelloGrayson, line 29: You're not incrementing the counts in the json reporter... this means summary will be empty

output/report.go
kriskowal, line 32: In this case, the copy for each pass-by-value might be heavy. @prashantv, do you have any advice on when to do pass-by-copy or pass-by-reference for structs?

kriskowal:
Consolidated in #59

output/report.go
prashantv, line 32: It does depend on the size of the config and how often you'll be calling the method. If you only ever call the method once for the whole process, then it's probably not worth optimizing. If this is something you'll call often, then it's probably worth using pointers.

The Go CodeReviewComments page has some info on this too:
https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

```
If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.
```

PR #57, opened 2016-05-23
Mux reporter

kriskowal:
Initial support for reporting both JSON and List output.

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,11 +52,19 @@ func main() {
 		os.Exit(1)
 	}
 
-	reporter.Start(config)
+	err = reporter.Start(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	for test := range results {
 		reporter.Next(test, config)
 	}
-	summary := reporter.End(config)
+	summary, err := reporter.End(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	output.Summarize(summary)
 
 	if summary.Failed == true {
--- a/main.go
+++ b/main.go
@@ -51,7 +51,12 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(config, results)
+
+	reporter.Start(config)
+	for test := range results {
+		reporter.Next(test, config)
+	}
+	summary := reporter.End(config)
 	output.Summarize(summary)
 
 	if summary.Failed == true {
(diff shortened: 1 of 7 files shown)


kriskowal:
Intent to rebase after #54, #56 land

output/list.go
```diff
@@ -32,26 +33,33 @@ var green = color.New(color.FgGreen).SprintFunc()
...
+	return nil
+}
+
+func (l *List) Next(test execute.TestResponse, config *plan.Config) {
+	for _, result := range test.Results {
+		var statStr string
+		switch result.Status {
+		case execute.Success:
+			statStr = green("✓")
+			l.Summary.NumSuccess++
```
HelloGrayson, line 50: Would love to pull the book-keeping for summary out of the ReporterFunc - the way it is right now is that every reporter has to do this, which is lame.

output/report.go
```diff
@@ -24,31 +24,37 @@ import (
...
 	// else a specific value was provided
 	switch name {
 	case "list":
-		return List, nil
+		return &List{}, nil
+	case "json":
+		return &JSON{}, nil
+	case "list,json":
+		fallthrough
+	case "json,list":
```
HelloGrayson: Yikes - hardcoding these?

kriskowal: Yeah, as an expedient. Maybe a temporary expedient, but need to give TTL attention before making this nice.

output/list.go
kriskowal, line 50: #58

output/report.go
```diff
@@ -29,29 +29,32 @@ import (
...
 	switch name {
 	case "list":
-		return List, nil
+		return &List{}, nil
 	case "json":
-		return JSON, nil
+		return &JSON{}, nil
+	case "list,json":
+		fallthrough
+	case "json,list":
```
prashantv, line 51: nit: instead of `fallthrough`, you can do

``` go
case "list,json", "json,list":
```

output/mux.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type Mux struct {
+	s         Summary
+	Reporters []Reporter
+}
+
+func (r Mux) Start(config *plan.Config) error {
+	for _, reporter := range r.Reporters {
+		err := reporter.Start(config)
```
prashantv, line 15: nit: inline into `if` below

main.go
```diff
@@ -51,7 +51,20 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(config, results)
+
+	err = reporter.Start(config)
+	if err != nil {
```
prashantv, line 56: nit: use `err := reporter.Start` and inline into if. that way the `err` is scoped only to the body of the `if`

kriskowal:
Consolidated in #59

output/report.go
kriskowal, line 51: neat

PR #59, opened 2016-05-23
Omnibus diff for multiple reporters

kriskowal:
Turns the summary into a reporter instance and factors it out of the JSON/List reporters. Includes the refactor from stream to {Start, Next, End} synchronous reporters.

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,21 +46,23 @@ func main() {
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	reporter, err := output.GetReporter(config.Reports)
-	if err != nil {
+	fail := func(err error) {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 
+	reporter, err := output.GetReporter(config.Reports)
+	if err != nil {
+		fail(err)
+	}
+
 	if err = reporter.Start(config); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		fail(err)
 	}
 	for test := range results {
 		reporter.Next(test)
 	}
 	if err := reporter.End(); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		fail(err)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -52,11 +52,19 @@ func main() {
 		os.Exit(1)
 	}
 
-	reporter.Start(config)
+	err = reporter.Start(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	for test := range results {
 		reporter.Next(test, config)
 	}
-	summary := reporter.End(config)
+	summary, err := reporter.End(config)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
 	output.Summarize(summary)
 
 	if summary.Failed == true {
--- a/main.go
+++ b/main.go
@@ -46,28 +46,21 @@ func main() {
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	reporter, err := output.GetReporter(config.Report)
+	reporter, err := output.GetReporter(config.Reports)
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 
-	err = reporter.Start(config)
-	if err != nil {
+	if err = reporter.Start(config); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 	for test := range results {
-		reporter.Next(test, config)
+		reporter.Next(test)
 	}
-	summary, err := reporter.End(config)
-	if err != nil {
+	if err := reporter.End(); err != nil {
 		fmt.Println(err)
-		os.Exit(1)
-	}
-	output.Summarize(summary)
-
-	if summary.Failed == true {
 		os.Exit(1)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -51,7 +51,12 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(config, results)
+
+	reporter.Start(config)
+	for test := range results {
+		reporter.Next(test, config)
+	}
+	summary := reporter.End(config)
 	output.Summarize(summary)
 
 	if summary.Failed == true {
--- a/output/json.go
+++ b/output/json.go
@@ -24,40 +24,26 @@ type JSONReport struct {
 	Behaviors map[string]*JSONBehaviorReport `json:"behaviors"`
 }
 
-var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
-	summary := Summary{}
-	report := JSONReport{
-		Behaviors: make(map[string]*JSONBehaviorReport),
-	}
+type JSON struct {
+	s Summary
+	r JSONReport
+}
+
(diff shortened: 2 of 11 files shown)


main.go
```diff
@@ -46,15 +46,21 @@ func main() {
...
-	reporter, err := output.GetReporter(config.Report)
+	reporter, err := output.GetReporter(config.Reports)
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	summary := reporter.Stream(config, results)
-	output.Summarize(summary)
 
-	if summary.Failed == true {
```
kriskowal, line 57: The failed case gets rolled up with summary.End()’s error condition.

main.go
```diff
@@ -46,15 +46,21 @@ func main() {
...
+	if err = reporter.Start(config); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
HelloGrayson: nit - perhaps we should move these 2 lines to a `Fail` function than can be used in the few cases below.

output/json.go
```diff
@@ -24,48 +25,58 @@ type JSONReport struct {
...
+	r    JSONReport
+	path string
+}
+
+func (j *JSON) Start(config *plan.Config) error {
+	j.path = config.JSONReportPath
+	j.r.Behaviors = make(map[string]*JSONBehaviorReport)
+
+	if config.JSONReportPath == "" {
+		return fmt.Errorf("JSON_REPORT_PATH is a required environment varialbe for REPORT=json")
```
HelloGrayson: nit - this error message probably shouldnt be coupled to the fact that the config came from env vars

output/json.go
```diff
@@ -24,48 +25,58 @@ type JSONReport struct {
 	Behaviors map[string]*JSONBehaviorReport `json:"behaviors"`
 }
 
-var JSON ReporterFunc = func(config *plan.Config, tests <-chan execute.TestResponse) Summary {
-	summary := Summary{}
-	report := JSONReport{
-		Behaviors: make(map[string]*JSONBehaviorReport),
+type JSON struct {
+	r    JSONReport
```
HelloGrayson: I'm not sure the brevity is worth it here - seeing `j.r...` below is a bit confusing. Suggest `j.report...` instead.

output/summary.go
```diff
@@ -31,22 +36,41 @@ type Summary struct {
 	NumSkipped int
 }
 
+func (s *Summary) Start(config *plan.Config) error {
```
HelloGrayson, line 39: nit - suggest passing the config in when creating the reporter struct, instead of passing it in here.

main.go
abhinav: Or just do `log.Fatal`.

output/json.go
abhinav: > varialbe

HelloGrayson:
Looking good 🐑

output/mux.go
```diff
@@ -0,0 +1,34 @@
+package output
+
+import (
+	"github.com/yarpc/crossdock/execute"
+	"github.com/yarpc/crossdock/plan"
+)
+
+type Mux struct {
+	Reporters []Reporter
```
abhinav: Optional: Can just do,

``` go
type Mux []Reporter

func (r Mux) Start(...)  error {
  for _, reporter := range r {
    // ...
  }
}
```

This will allow usage like so,

``` go
reporter := output.Mux{a, b, c}
```

output/list.go
```diff
@@ -33,26 +33,28 @@ var green = color.New(color.FgGreen).SprintFunc()
...
-				statStr = yellow("S")
-				summary.NumSkipped++
-			default:
-				statStr = red("F")
-				summary.Failed = true
-				summary.NumFail++
-			}
-			fmt.Printf("%v - %v - %v\n", statStr, test.TestCase, result.Output)
+type List struct {
+}
```
abhinav: really nitty nit: close empty structs on the same line with no space between the keyword and the `{`.

``` go
type List struct{}
```

---

Also this is very optional, mostly an FYI: For stateless implementations of an interface, it's acceptable to provide a global instance:

``` go
type list struct{}

func (l *list) ...

var List Reporter = &list{}
```

Then you can just use `output.List` as the reporter instead of `&output.List{}`.

output/list.go
```diff
@@ -33,26 +33,28 @@ var green = color.New(color.FgGreen).SprintFunc()
...
+
+func (l *List) Start(config *plan.Config) error {
+	return nil
```
abhinav, line 41: If you're not actually using the receiver, you can do,

``` go
func (*List) Start(config *plan.Config) error {
  ..
}
```

output/list.go
```diff
@@ -33,26 +33,28 @@ var green = color.New(color.FgGreen).SprintFunc()
...
+}
+
+func (l *List) Next(test execute.TestResponse) {
```
abhinav: Same. No need to name the receiver if you're not using it.

Also, making it a pointer doesn't serve any value here either. You can just use `func (List) ...`.

main.go
HelloGrayson: I'm not using `log.Fatal` intentionally here so I have full control over the output

output/report.go
```diff
@@ -29,30 +29,29 @@ import (
...
 	}
-	// else a specific value was provided
-	switch name {
-	case "list":
-		return List, nil
-	case "json":
-		return JSON, nil
-	default:
-		return nil, fmt.Errorf("%v is not a valid reporter", name)
+	reporters := []Reporter{}
```
abhinav: nit: for empty slices, you want to use,

``` go
var reporters []Reporter
```

`nil` is a valid empty slice.

See also https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

output/report.go
```diff
@@ -29,30 +29,29 @@ import (
...
 }
 
 // GetReporter returns a ReporterFunc for the given name
-func GetReporter(name string) (Reporter, error) {
+func GetReporter(names []string) (Reporter, error) {
 	// default reporter
-	if name == "" {
-		return List, nil
+	if len(names) == 0 {
+		names = append(names, "list")
```
abhinav: Why not short circuit early:

```go
if len(names) == 0 {
  return &List{}, nil
}
```

output/list.go
HelloGrayson: I've always thought a better default was to use pointer receivers... at least that way you don't get some new Go person (like me) come along and get confused when they cant modify data on the struct :P

output/report.go
kriskowal: It’s a little more involved since the default is a `Mux{Reporters: []Reporter{&List{}, &Summary{}}`. I think it’s cleaner to let it do its thing.

output/list.go
abhinav: Discussed offline but: Yeah, always-pointers is more newb friendly but pointers-for-mutation-and-large-structs is more correct. :)

output/report.go
abhinav: Oh, I didn't realize the default is still muxed. Makes sense. Deleting comment.

output/json.go
kriskowal: On the other hand, the environment variable makes it more clear to the end user how to mitigate the problem. Would you alternately recommend handling the error in main to surface the user facing message?

HelloGrayson: For now, I'd say lets leave it until we allow multiple config sources, eg env and yaml file

abhinav:
LGTM

HelloGrayson:
🎈 🎈 🎈

PR #65, opened 2016-06-20
Better timeout logic & messages.

bombela:
The crossdock driver "HTTP PING" every other containers before running
the tests.

This adds a smaller sub-timeout of 500ms, making sure that we cancel and
retry the HTTP HEAD request from scratch during the global wait timeout.

Additional this fixes few log messages.

### DIFF
--- a/execute/wait.go
+++ b/execute/wait.go
@@ -80,8 +80,8 @@ func waitForHTTPRequest(host string, cancel <-chan struct{}) {
 		go func() {
 			resp, err := client.Do(req)
 			if err == nil && resp.StatusCode != http.StatusOK {
-				err = errors.New(fmt.Sprintf("Expecting %v: got %v",
-					http.StatusOK, resp.Status))
+				err = fmt.Errorf("expecting %v, got %v",
+					http.StatusOK, resp.Status)
 			}
 			c <- err
 		}()
--- a/execute/wait.go
+++ b/execute/wait.go
@@ -51,7 +51,7 @@ func Wait(hosts []string, timeout time.Duration) {
 	wg.Wait()
 
 	if !timer.Stop() {
-		log.Fatalf("Error: One or more services timed out after %d second(s)", timeout)
+		log.Fatalf("Error: One or more services timed out after %v", timeout)
 	}
 	fmt.Printf("\nAll services are up after %v!\n", time.Since(begin))
 }
@@ -66,15 +66,16 @@ func waitForHTTPRequest(host string, cancel <-chan struct{}) {
 
 	err := errors.New("init")
 	for err != nil {
+		log.Printf("HTTP: HEAD %s\n", url.String())
 		req, reqErr := http.NewRequest("HEAD", url.String(), nil)
 		if reqErr != nil {
 			log.Printf("Warning: Failed to create request for URL '%s' -  skipping service '%s'",
 				url.String(), host)
 			return
 		}
+		req.Cancel = cancel
 
-		tr := &http.Transport{}
-		client := &http.Client{Transport: tr}
+		client := &http.Client{Transport: &http.Transport{}, Timeout: 500 * time.Millisecond}
 		c := make(chan error, 1)
 		go func() {
 			_, err := client.Do(req)
@@ -83,7 +84,6 @@ func waitForHTTPRequest(host string, cancel <-chan struct{}) {
 
 		select {
 		case <-cancel:
-			tr.CancelRequest(req)
 			log.Printf("HTTP: Service %v timed out. Last error: %v", host, err)
 			<-c
 			return
--- a/execute/wait.go
+++ b/execute/wait.go
@@ -78,7 +78,11 @@ func waitForHTTPRequest(host string, cancel <-chan struct{}) {
 		client := &http.Client{Transport: &http.Transport{}, Timeout: 500 * time.Millisecond}
 		c := make(chan error, 1)
 		go func() {
-			_, err := client.Do(req)
+			resp, err := client.Do(req)
+			if err == nil && resp.StatusCode != http.StatusOK {
+				err = errors.New(fmt.Sprintf("Expecting %v: got %v",
+					http.StatusOK, resp.Status))
+			}
 			c <- err
 		}()
 
--- a/execute/wait_test.go
+++ b/execute/wait_test.go
@@ -61,7 +61,7 @@ func TestObvious(t *testing.T) {
 	defer ts.Close()
 	host := ts.Listener.Addr().String()
 	cancel := make(chan struct{})
-	timer := time.AfterFunc(time.Second*2, func() {
+	timer := time.AfterFunc(2*time.Second, func() {
 		close(cancel)
 	})
 	waitForHTTPRequest(host, cancel)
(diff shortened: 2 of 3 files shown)


execute/wait.go
```diff
@@ -51,7 +51,7 @@ func Wait(hosts []string, timeout time.Duration) {
 	wg.Wait()
 
 	if !timer.Stop() {
-		log.Fatalf("Error: One or more services timed out after %d second(s)", timeout)
+		log.Fatalf("Error: One or more services timed out after %v", timeout)
```
HelloGrayson, line 54: Does `%v` here indicate the unit? What's an example of how this would look?

HelloGrayson:
How did you test this?

execute/wait.go
abhinav, line 54: Yeah, `time.Duration`'s default formatting includes the unit: https://play.golang.org/p/49k0GBiVS-

bombela:
Easy, don't start the go container and see how many times it retries vs no-retry.

abhinav:
@bombela It would be nice to have some integration tests for this functionality. It may not be fully straightforward to do that with the code as-is though, especially because all the timeouts are hardcoded. If you think you can add an integration test with this as-is, go for it. Otherwise we can merge this change and make a separate one with an integration test, performing the necessary changes to the code in that.

main.go
```diff
@@ -32,7 +33,7 @@ import (
 )
 
 func main() {
-	fmt.Printf("\nCrossdock starting...\n\n")
+	fmt.Printf("\nCrossdock starting (go %v)...\n\n", runtime.Version())
```
prashantv: I'm not sure if this belongs on the startup print. Of all the information in the crossdock environment that might be useful, I'm not sure the Go version is the most important.

While you were debugging your specific issue, I'm sure this came in handy, but do you expect many problems where this will come in useful?

HelloGrayson: Agreed - I want the fact that Crossdock is go to be transparant to the user.

bombela: As you wish. It was tricky to debug what version of go was running in the container, and I am not sure anybody would care about the go version in the output until you have to troubleshoot. :+1:

HelloGrayson:
:shipit: once travis is happy :)

bombela:
Please review now with some unit testing and the HTTP PING also checks for status 200.

execute/wait.go
```diff
@@ -66,24 +66,28 @@ func waitForHTTPRequest(host string, cancel <-chan struct{}) {
...
-		tr := &http.Transport{}
-		client := &http.Client{Transport: tr}
+		client := &http.Client{Transport: &http.Transport{}, Timeout: 500 * time.Millisecond}
 		c := make(chan error, 1)
 		go func() {
-			_, err := client.Do(req)
+			resp, err := client.Do(req)
+			if err == nil && resp.StatusCode != http.StatusOK {
+				err = errors.New(fmt.Sprintf("Expecting %v: got %v",
+					http.StatusOK, resp.Status))
```
abhinav: Use `fmt.Errorf` instead

```
err = fmt.Errorf("expected HTTP status %v but got %v", ...)
```

Also, error strings usually start with lower case so that you can combine them with ':'. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	host := ts.Listener.Addr().String()
+	cancel := make(chan struct{})
+	timer := time.AfterFunc(time.Second*2, func() {
+		close(cancel)
+	})
+	waitForHTTPRequest(host, cancel)
+	require.False(t, timer.Stop())
+}
+
+func TestNotReadyThenOk(t *testing.T) {
```
HelloGrayson, line 86: naming++ :)

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+type testHandler struct {
+	t             *testing.T
+	delay         time.Duration // delay before replying
+	notReadyCount int32         // 500 that many time before returning success.
+}
+
+func (h *testHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	t := h.t
+	if h.delay != 0 {
+		t.Logf("serverHTTP: Delaying for %v", h.delay)
```
abhinav: serveHTTP, not serverHTTP. Although I would name the type here instead ("testHandler")

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if r.Method == "HEAD" {
+		w.Header().Add("Content-Length", "0")
+		t.Logf("serverHTTP: Returning %v", http.StatusOK)
+		return
+	}
+	http.NotFound(w, r)
+}
+
+func TestObvious(t *testing.T) {
+	ts := httptest.NewServer(&testHandler{t: t, delay: 0 * time.Second})
```
abhinav: Don't need to specify delay if it's 0. Zero value for it is already 0.

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	defer ts.Close()
+	host := ts.Listener.Addr().String()
```
abhinav, line 62: You can use `ts.URL` instead

https://golang.org/pkg/net/http/httptest/#Server

execute/wait_test.go
bombela: I wanted to be explicit for some reason. Its now removed.

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestObvious(t *testing.T) {
+	ts := httptest.NewServer(&testHandler{t: t})
+	defer ts.Close()
+	host := ts.Listener.Addr().String()
+	cancel := make(chan struct{})
+	timer := time.AfterFunc(time.Second*2, func() {
+		close(cancel)
+	})
+	waitForHTTPRequest(host, cancel)
+	require.True(t, timer.Stop())
```
abhinav: I would just use `defer timer.Stop()` instead. Nothing to worry about if it returns false.

execute/wait_test.go
bombela, line 62: Depends if we can change the signature of the wait function...

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	http.NotFound(w, r)
+}
+
+func TestObvious(t *testing.T) {
+	ts := httptest.NewServer(&testHandler{t: t})
+	defer ts.Close()
+	host := ts.Listener.Addr().String()
+	cancel := make(chan struct{})
+	timer := time.AfterFunc(time.Second*2, func() {
```
abhinav: Prefer `2*time.Second` so that it reads as '2 seconds'

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+	waitForHTTPRequest(host, cancel)
+	require.True(t, timer.Stop())
+}
+
+func TestTooSlow(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	ts := httptest.NewServer(&testHandler{t: t, delay: 1 * time.Second})
```
abhinav: Use just `time.Second` for '1 second'

execute/wait_test.go
abhinav: consider adding a message in front:

```
require.True(t, timer.Stop(), "request should not run for 2 seconds")
```

execute/wait_test.go
```diff
@@ -0,0 +1,99 @@
...
+	defer ts.Close()
+	host := ts.Listener.Addr().String()
+	cancel := make(chan struct{})
+	timer := time.AfterFunc(time.Second*2, func() {
+		close(cancel)
+	})
+	waitForHTTPRequest(host, cancel)
+	require.False(t, timer.Stop())
```
abhinav: Same here,

```
require.False(t, timer.Stop(), "request should take over 2 seconds")
```

abhinav:
Looks good with nits.

abhinav:
LGTM

PR #66, opened 2016-06-28
Finish moving to Crossdock org

HelloGrayson:
@abhinav @kriskowal @bombela


yurishkuro:
Where will reusable things like sink, behavior funcs go?

HelloGrayson:
@yurishkuro check out github.com/crossdock/crossdock-go

HelloGrayson:
@yurishkuro I'm moving everything to this org so we have room to grow, i.e. add additional clients

HelloGrayson:
(abg says ship it)

yurishkuro:
crossdock-go is a bit confusing, e.g. it has `combinations.go`, which is not something that behavior implementation needs, and `doc.go` refers to `behavior.Sink` which I don't see. Maybe it's WIP?

HelloGrayson:
@yurishkuro its def a WIP. Take a look at these files in yarpc-go:
- how to setup crossdock -  https://github.com/yarpc/yarpc-go/blob/master/crossdock/client/start.go
- how to integrate crossdock into regular test suite - https://github.com/yarpc/yarpc-go/blob/master/crossdock/main_test.go

You'll see that `crossdock.Combinations` is being used in the second one.

yurishkuro:
> You'll see that crossdock.Combinations is being used in the second one.

I see, so it essentially runs the crossdock-style test against a single repo without docker. Neat. Although, it looks like it's using the same fixed ports as the real crossdock test, does it not cause issues in CI?

HelloGrayson:
@yurishkuro hasn't been a problem yet, as we run these serially. 

This lib has a long way to go and there is a ton of things we can do to improve the experience - please create issues as you see things!

PR #68, opened 2016-07-01
Stable order of test execution

bombela:
- Configuration is properly sorted.
- Test plan is generated in a stable manner.

This means tests are always running in the same order for a given configuration.

### DIFF
--- a/examples/guide/main.go
+++ b/examples/guide/main.go
@@ -32,7 +32,6 @@ func main() {
 
 		// when method is HEAD, report back with a 200 when ready to run tests
 		if r.Method == "HEAD" {
-			w.Header().Add("Content-Length", "0")
 			return
 		}
 
--- a/examples/guide/main.go
+++ b/examples/guide/main.go
@@ -30,9 +30,9 @@ func main() {
 		// http://<test-client>:8080/?behavior=dance
 		behavior := r.FormValue("behavior")
 
-		// when client is called with no arguments,
-		// report back with a 200 when ready to run tests
-		if behavior == "" {
+		// when method is HEAD, report back with a 200 when ready to run tests
+		if r.Method == "HEAD" {
+			w.Header().Add("Content-Length", "0")
 			return
 		}
 
--- a/execute/wait_test.go
+++ b/execute/wait_test.go
@@ -42,7 +42,7 @@ func (h *testHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		time.Sleep(h.delay)
 	}
 	if h.notReadyCount > 0 {
-		h.notReadyCount -= 1
+		h.notReadyCount--
 		t.Logf("testHandler: Returning %v (left: %v)",
 			http.StatusServiceUnavailable, h.notReadyCount)
 		http.Error(w, "Server not ready", http.StatusServiceUnavailable)
--- a/main.go
+++ b/main.go
@@ -23,7 +23,6 @@ package main
 import (
 	"fmt"
 	"log"
-	"os"
 	"time"
 
 	"github.com/crossdock/crossdock/execute"
@@ -38,30 +37,28 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	reporter, err := output.GetReporter(config.Reports)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	plan := plan.New(config)
 
-	fmt.Printf("Waiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
+	if err = reporter.Start(plan); err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("\nWaiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
 	execute.Wait(plan.Config.WaitForHosts, time.Duration(30)*time.Second)
 
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
 
-	fail := func(err error) {
-		os.Exit(1)
-	}
-
-	reporter, err := output.GetReporter(config.Reports)
-	if err != nil {
-		fail(err)
-	}
-
-	if err = reporter.Start(config); err != nil {
-		fail(err)
-	}
 	for test := range results {
 		reporter.Next(test)
 	}
 	if err := reporter.End(); err != nil {
-		fail(err)
+		log.Fatal(err)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -45,7 +45,7 @@ func main() {
 
 	plan := plan.New(config)
 
-	if err = reporter.Start(plan); err != nil {
+	if err := reporter.Start(plan); err != nil {
 		log.Fatal(err)
 	}
 
--- a/output/list.go
+++ b/output/list.go
@@ -35,7 +35,7 @@ var red = color.New(color.FgRed).SprintFunc()
 
 type list struct{}
 
-// List is the output reporter as a list.
+// List is a simple Reporter that prints the test output to stdout.
 var List list
 
 func (list) Start(plan *plan.Plan) error {
--- a/output/list.go
+++ b/output/list.go
@@ -35,6 +35,7 @@ var red = color.New(color.FgRed).SprintFunc()
 
 type list struct{}
 
+// List is the output reporter as a list.
 var List list
 
 func (list) Start(plan *plan.Plan) error {
--- a/output/list.go
+++ b/output/list.go
@@ -37,7 +37,7 @@ type list struct{}
 
 var List list
 
-func (list) Start(config *plan.Config) error {
+func (list) Start(plan *plan.Plan) error {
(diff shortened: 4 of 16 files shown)


examples/guide/main.go
```diff
@@ -30,9 +30,9 @@ func main() {
 		// http://<test-client>:8080/?behavior=dance
 		behavior := r.FormValue("behavior")
 
-		// when client is called with no arguments,
-		// report back with a 200 when ready to run tests
-		if behavior == "" {
+		// when method is HEAD, report back with a 200 when ready to run tests
+		if r.Method == "HEAD" {
+			w.Header().Add("Content-Length", "0")
```
abhinav: FYI this is needed so that certain HTTP clients know not to expect a body and disconnect afterwards.

plan/entities.go
```diff
@@ -41,47 +38,46 @@ type Axis struct {
...
 }
 
-// Behavior represents the test behavior will be triggered by crossdock
+type Axes []Axis
+
+func (a Axes) Len() int           { return len(a) }
+func (a Axes) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a Axes) Less(i, j int) bool { return a[i].Name < a[j].Name }
+
+func (a Axes) Indexed() map[string]Axis {
```
abhinav: Nit: rename to `Index()`

plan/entities.go
```diff
@@ -41,47 +38,46 @@ type Axis struct {
 	Values []string
 }
 
-// Behavior represents the test behavior will be triggered by crossdock
+type Axes []Axis
```
abhinav, line 42: docs over exported entities please

examples/guide/main.go
prashantv: can we put a comment in the code for this, and mention what clients

kriskowal:
:ok:

prashantv:
nice, lgtm

execute/wait_test.go
```diff
@@ -42,7 +42,7 @@ func (h *testHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		time.Sleep(h.delay)
 	}
 	if h.notReadyCount > 0 {
-		h.notReadyCount -= 1
+		h.notReadyCount--
```
HelloGrayson, line 45: oh how I missed c-style langs :)

main.go
```diff
@@ -38,30 +37,28 @@ func main() {
...
-
-	if err = reporter.Start(config); err != nil {
-		fail(err)
-	}
 	for test := range results {
 		reporter.Next(test)
 	}
 	if err := reporter.End(); err != nil {
-		fail(err)
+		log.Fatal(err)
```
HelloGrayson, line 62: Please don't log, that adds the timestamp to the output and gives the reporter less control

plan/config.go
```diff
@@ -83,10 +86,12 @@ func parseBehavior(d string) Behavior {
 	key := strings.ToLower(pair[0])
 	values := strings.Split(pair[1], ",")
 	values = trimCollection(values)
+	sort.Strings(values)
```
breerly: nice, ty sir

plan/new_test.go
```diff
@@ -26,201 +26,157 @@ import (
...
 		},
-		plan.TestCases,
-	)
+	}
+	for _, tt := range tests {
+		assert.Equal(t, tt.o, combinations(tt.i))
+	}
+}
+
+func TestNew(t *testing.T) {
```
HelloGrayson, line 113: this is a great test 👍

HelloGrayson:
lgtm

execute/wait_test.go
bombela, line 45: I find -= 1 so much more readable. Its a matter of taste, and golint doesn't agree with me obviously ;)

main.go
bombela, line 62: ok, so here, reporter.End() should not fail. It it fails, its an error in crossdock, do we really want to hide the error here?

Or is it in the contract of reporters, that if a reporter function returns an error, it means "I have printed errors shit, now please exit". Because in this case, I would argue that this is not clear at all, and we should return (status, err). status being an enum of what action to perform (continue or exit).

Any example why a reporting function would want to abort the whole crossdock by purpose?

plan/new_test.go
bombela, line 113: thank @abhinav for it, not me!

main.go
HelloGrayson, line 62: I thought this was for if the test suite failed. Ok, this is probably fine then 👍

output/list.go
```diff
@@ -35,9 +35,10 @@ var red = color.New(color.FgRed).SprintFunc()
 
 type list struct{}
 
+// List is the output reporter as a list.
```
abhinav: Change this to something like,

> List is a simple Reporter that prints the test output to stdout.

plan/entities.go
```diff
@@ -41,47 +38,49 @@ type Axis struct {
 	Values []string
 }
 
-// Behavior represents the test behavior will be triggered by crossdock
+// Axes is a slice of "Axis"
```
abhinav, line 41: > Axes is a collection of Axis objects sortable by axis name.

plan/entities.go
```diff
@@ -41,47 +38,49 @@ type Axis struct {
...
 type Behavior struct {
-	Name    string
-	Clients string
-	Params  []string
+	Name       string
+	ClientAxis string
+	ParamsAxes []string
 }
 
+// Behaviors is a sortable slice of "Behavior".
```
abhinav: > Behaviors is a collection of Behavior objects sortable by behavior name.

abhinav:
LGTM with nits

examples/guide/main.go
bombela: technically any confirming client for http>=1.1

Go keeps the TCP connection open, but doesn't (cannot?) enforce that you have followed http properly and informed the client that there is nothing else to read on this request.

Obviously, http2 with its explicit streams is better off :)

prashantv: A `HEAD` request never has a body -- by definition of how HEAD is defined:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4

```
The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. 
```

It looks like the `Content-Length` header should be the size of the body that would have been returned if this was a `GET`. In this case, since we don't know how many bytes would have been returned, not including a `Content-Length` header at all seems to be the correct response.

See related:
http://stackoverflow.com/questions/3854842/content-length-header-with-head-requests

I verified that `google.com` doesn't return a `Content-Length` for a `HEAD` request to `/`:

```
> telnet www.google.com 80
Trying 216.58.195.228...
Connected to www.google.com.
Escape character is '^]'.
HEAD / HTTP/1.1
Host: www.google.com

HTTP/1.1 200 OK
Date: Wed, 06 Jul 2016 17:28:40 GMT
Expires: -1
Cache-Control: private, max-age=0
Content-Type: text/html; charset=ISO-8859-1
P3P: CP="This is not a P3P policy! See https://www.google.com/support/accounts/answer/151657?hl=en for more info."
Server: gws
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Set-Cookie: NID=81=wiXmbp5h9Blkq0xR1gRrz5T_B49qL2I4m00E69i6e-wnmH_l9rWrC_mPGTJ_n7JfawvcmtyWHhkLG_nc1b7mn9UHIUCEizEoLBoIHzmi6Bsp5p2PWfv09RshimyY2YjXil_Iz3qF6ALiCQ; expires=Thu, 05-Jan-2017 17:28:40 GMT; path=/; domain=.google.com; HttpOnly
Transfer-Encoding: chunked
Accept-Ranges: none
Vary: Accept-Encoding
```

So I'm very suspicious that this is required for our use case. Let's discuss offline to make sure the issue isn't somewhere else.

bombela: I stand corrected. I was tricked by curl -X HEAD vs curl --head :) In the former, curl doesn't understand the request as a HEAD request, but as a generic http request. Go http server is smart enough to fill the content-length when required.

main.go
```diff
@@ -38,30 +37,28 @@ func main() {
...
+
+	reporter, err := output.GetReporter(config.Reports)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	plan := plan.New(config)
 
-	fmt.Printf("Waiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
+	if err = reporter.Start(plan); err != nil {
```
prashantv: nit: use `err :=` here. where possible, you want the variable to be scoped as narrowly as possible

plan/entities.go
prashantv, line 41: +1

prashantv:
lgtm

execute/wait_test.go
HelloGrayson, line 45: Must be the PHP in me :P

PR #69, opened 2016-07-01
Pretty reporting outputs.

bombela:
`compact` is a new reporting output. a compact form of the list output.

The summary output also prints the current configuration.

### DIFF
--- a/execute/entities.go
+++ b/execute/entities.go
@@ -33,7 +33,8 @@ type TestResponse struct {
 	Results  []Result
 }
 
-// StatusSummary contains the a status report of the TestReponse, see SummarizeStatus.
+// StatusSummary contains the status report of a TestReponse, see
+// SummarizeStatus.
 type StatusSummary struct {
 	Status  Status
 	Total   int
--- a/execute/entities.go
+++ b/execute/entities.go
@@ -33,6 +33,47 @@ type TestResponse struct {
 	Results  []Result
 }
 
+// StatusSummary contains the a status report of the TestReponse, see SummarizeStatus.
+type StatusSummary struct {
+	Status  Status
+	Total   int
+	Skipped int
+	Passed  int
+	Failed  int
+}
+
+// SummarizeStatus returns a summary of the TestResponse, compiling it from all
+// the Results pertaining to it.
+func (test *TestResponse) SummarizeStatus() StatusSummary {
+	total := len(test.Results)
+	passed := 0
+	skipped := 0
+	for _, result := range test.Results {
+		switch result.Status {
+		case Success:
+			passed++
+		case Skipped:
+			skipped++
+		}
+	}
+	failed := total - passed - skipped
+
+	behaviorStatus := Failed
+	if skipped == total {
+		behaviorStatus = Skipped
+	} else if failed == 0 {
+		behaviorStatus = Success
+	}
+
+	return StatusSummary{
+		Status:  behaviorStatus,
+		Total:   total,
+		Skipped: skipped,
+		Passed:  passed,
+		Failed:  failed,
+	}
+}
+
 // Result is the outcome of an individual test case ran by the test client
 type Result struct {
 	Status Status
--- a/output/list.go
+++ b/output/list.go
@@ -22,6 +22,8 @@ package output
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/crossdock/crossdock/execute"
 	"github.com/crossdock/crossdock/plan"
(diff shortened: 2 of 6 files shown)


HelloGrayson:
@bombela can you run this against yarpc-go and post the output? Should be able to switch the Crossdock image in yarpc-go's docker-compose.yml to this branch name:

```
services:
    crossdock:
        image: crossdock/crossdock:reporting
        ...
```

bombela:
An extract of the list reporting:
list

An the full output of the compact version:
compact

kriskowal:
LG™

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
 var green = color.New(color.FgGreen).SprintFunc()
 var yellow = color.New(color.FgYellow).SprintFunc()
 var red = color.New(color.FgRed).SprintFunc()
+var blue = color.New(color.FgBlue).SprintFunc()
```
prashantv: nit: good place to use a group

``` go
var (
  green = color.New(...)
  yellow = color.New(...)
  [...]
)
```

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
+
+func (list *List) Start(plan *plan.Plan) error {
 	return nil
 }
 
-func (list) Next(test execute.TestResponse) {
+func statusToColoredSymbol(status execute.Status) string {
+	switch status {
+	case execute.Success:
+		return green("✓")
```
prashantv: since it's not printing anything, is there any point in calling `green` here? could you just return `""`

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
+	return fmt.Printf(format, a...)
+}
 
-func (list) Start(plan *plan.Plan) error {
+func (list *List) printDot(status execute.Status) {
+	list.dotted = true
+	var dot string
+	switch status {
+	case execute.Skipped:
+		dot = yellow(".")
```
prashantv: nit: maybe a `-` for skipped? something slightly different might be easier to see than white vs yellow

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
+		} else {
+			list.printf("%v %v ⇒ %v\n", statusToColoredSymbol(result.Status),
+				fmtTestCase(test.TestCase), result.Output)
+		}
+		return
+	}
+
+	total := len(test.Results)
+	passed := 0
+	skipped := 0
```
prashantv: optionally: since you're using default values, you could do

``` go
var passed, skipped int
```

output/summary.go
```diff
@@ -38,6 +41,27 @@ type Summary struct {
 }
 
 func (s *Summary) Start(plan *plan.Plan) error {
+	config := plan.Config
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 8, 0, ' ', tabwriter.AlignRight)
+	defer w.Flush()
+
+	fmt.Fprintf(w, "Configuration:\n")
```
HelloGrayson, line 49: I would move the config output into it's own reporter, users choosing the summary output probably don't want to see this, and if they should be able to just:

```
REPORTER=config,summary
```

output/list.go
HelloGrayson: I would use `S` for skipped - I'm color-blind and honestly I didn't see that the dots were different colors.

output/report.go
```diff
@@ -44,7 +44,9 @@ func GetReporter(names []string) (Reporter, error) {
 	for _, name := range names {
 		switch name {
 		case "list":
-			mux = append(mux, List)
+			mux = append(mux, &List{Compact: false})
+		case "compact":
+			mux = append(mux, &List{Compact: true})
```
HelloGrayson: Let's make this an entirely separate reporter, instead of adding a flag to the list.

output/summary.go
```diff
@@ -57,16 +81,11 @@ func (s *Summary) Next(test execute.TestResponse) {
 
 func (s *Summary) End() error {
 	fmt.Println("")
+	total := s.NumSuccess + s.NumFail + s.NumSkipped
```
HelloGrayson, line 84: Let's move this to `Summary.Total` instead of calculating it here if possible. That way all reporters have access to it.

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
 	return nil
 }
 
-func (list) Next(test execute.TestResponse) {
+func statusToColoredSymbol(status execute.Status) string {
+	switch status {
+	case execute.Success:
+		return green("✓")
+	case execute.Skipped:
+		return yellow("⏩")
```
HelloGrayson: It's not obvious to me that this means skipped in the output - suggest a `S` or finding something else.

HelloGrayson:
I'd like to see the reporter renamed to `minimal` and the output to look something like #51

At a bare minimum we should:
- display each behavior in the output with minimal indicators for pass/success/fail
- summarize the totals once the test is done, then:
- list errors after the test summary

```
Starting Crossdock...

RAW .....
JSON .............
THRIFT .........................................
ERRORS .......................................................
HEADERS ...........F..........SS...........

100 passed
1 failed
2 skipped

F - HEADERS java -> server=go, transport=tchannel - <output>
```

output/list.go
bombela: what do you mean by its not printing anything? are you talking about this: `return green("✓")`

output/list.go
bombela: sounds good yellow("S") it is.

output/report.go
bombela: is it really necessary? because the compact and list reporters share almost everything. Sure, I can write a common lib but its kinda adding abstraction for the sake of it?

output/list.go
prashantv: hmm does Github have a bug? this is what I see:
image

output/summary.go
bombela, line 84: Reporters don't have access to the summary reporter.

output/list.go
bombela: "⏩" is "skip" on music players. but "S" it will be then.

bombela:
If we group by behaviors, we loose the ability to easily run tests in any random order.

I said easily because sure, we can buffer everything, and then print the output selectively. But this means no more interactive progress. Additionally I am not sure I want to use ncurses for a multiline interactive output either.

In my compact output, everything should be compact enough that you do not need additional summary of errors.

output/list.go
bombela: ah fun! no problem on github for me (chrome on linux), I see the unicode checkmark.

output/report.go
HelloGrayson: We should optimize for:
- Building new reporters - if a third reporter wants to come in and use some of these pieces, the example this sets fourth is to come into `List` and make it do even more. A common lib would be better.
- Single responsibility - a reporter should do 1 thing well, and that thing is show one version of output. A flag that changes its behavior breaks this idea.
- Isolation - a new reporter shouldn't change another one. In this case, instead of just a new reporter, we're changing `List` significantly.

HelloGrayson:
We shouldn't consider random order or concurrent tests yet - there are creative ways we can solve that without ncurses (docker binaries do this). Let's solve that when we get there and for now focus on the best possible output without considering those variables.

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
+func (list *List) printDot(status execute.Status) {
+	list.dotted = true
+	var dot string
+	switch status {
+	case execute.Skipped:
+		dot = yellow(".")
+	default:
+		dot = "."
+	}
+	fmt.Printf(dot)
```
abhinav: If you're not doing any formatting, `fmt.Print(dot)` should suffice.

output/list.go
```diff
@@ -32,31 +34,127 @@ import (
...
 
-func (list) Start(plan *plan.Plan) error {
+func (list *List) printDot(status execute.Status) {
+	list.dotted = true
+	var dot string
+	switch status {
+	case execute.Skipped:
+		dot = yellow(".")
+	default:
+		dot = "."
```
abhinav: Nit: If there's only one condition, a simple if might be better:

``` go
dot := "."
if status == execute.Skipped {
  dot = yellow("S")
}
```

output/list.go
abhinav: lol apparently that happens only in split mode. works just fine in unified diff mode.

abhinav:
Besides mine and others' nits, I'm okay with the the two output options as they are. But I agree with @breerly that we should split the reporter into two rather than adding flags on the existing one.

@breerly, does this sound good? We can iterate on the output of the compact reporter later if necessary.

output/summary.go
abhinav, line 49: Alternatively, if this is trivially doable later, make a GH issue and defer.

output/report.go
bombela: I think keeping things simple is more important. But for the scope, it doesn't matter much, and I will split into two distinct reporters.

output/summary.go
HelloGrayson, line 84: Fair 👍

HelloGrayson:
SGTM!

HelloGrayson:
Let's create issues for non-blocking parts, split the reporters, and ship it :)

bombela:
stuff is split

execute/entities.go
```diff
@@ -33,6 +33,48 @@ type TestResponse struct {
 	Results  []Result
 }
 
+// StatusSummary contains the status report of a TestReponse, see
```
HelloGrayson, line 36: TestResponse*

execute/entities.go
```diff
@@ -33,6 +33,48 @@ type TestResponse struct {
...
+	Status  Status
+	Total   int
+	Skipped int
+	Passed  int
+	Failed  int
+}
+
+// SummarizeStatus returns a summary of the TestResponse, compiling it from all
+// the Results pertaining to it.
+func (test *TestResponse) SummarizeStatus() StatusSummary {
```
HelloGrayson, line 48: nice, this feels like a much better place for this

HelloGrayson:
:shipit:

execute/entities.go
```diff
@@ -33,6 +33,48 @@ type TestResponse struct {
...
+			skipped++
+		}
+	}
+	failed := total - passed - skipped
+
+	behaviorStatus := Failed
+	if skipped == total {
+		behaviorStatus = Skipped
+	} else if failed == 0 {
+		behaviorStatus = Success
```
abhinav, line 66: optional nit: You can use a switch statement here.

``` go
switch {
case skipped == total:
  behaviorStatus = Skipped
case failed == 0:
  behaviorStatus = Success
}
```

abhinav:
LGTM

PR #74, opened 2016-07-12
Fail test suite when client returns 0 results

### LINKED ISSUE #70: No results from a client should fail
Currently it's possible for a client to return `[]` as a response, and it won't fail. I noticed this in @bombela's PR https://github.com/crossdock/crossdock/pull/69. When run from yarpc-go with `REPORT=compact`, we get see the following 4 tests don't return a single subresult:

```
∅ [errorstchout] go→ (errorstchout=go server=go) ⇒ (no results)
∅ [errorstchout] go→ (errorstchout=go server=java) ⇒ (no results)
∅ [errorstchout] go→ (errorstchout=go server=node) ⇒ (no results)
∅ [errorstchout] go→ (errorstchout=go server=python) ⇒ (no results)
```

The culprit looks like https://github.com/crossdock/crossdock/blob/master/execute/run.go#L113:

``` go
func toResults(subResponses []subResponse) []Result {
    var results []Result
    for _, subResponse := range subResponses {
        status := Failed
        switch subResponse.Status {
        case "passed":
            status = Success
        case "skipped":
            status = Skipped
        }
        result := Result{
            Status: status,
            Output: subResponse.Output,
        }
        results = append(results, result)
    }
    return results
}
```

If `len(subResponses) == 0`, then we should add a failed `Result` to `results` and return.

This way clients will be required to run at least 1 test for every behavior provided, or skip explicitly.

cc @kriskowal this will cause those tests in yarpc-go to begin failing

HelloGrayson:
fixes #70

### DIFF
--- a/execute/run.go
+++ b/execute/run.go
@@ -68,6 +68,16 @@ func executeTestCase(testCase plan.TestCase) TestResponse {
 		}
 	}
 
+	if len(subResponses) == 0 {
+		return TestResponse{
+			TestCase: testCase,
+			Results: []Result{{
+				Status: Failed,
+				Output: "client returned 0 results",
+			}},
+		}
+	}
+
 	return TestResponse{
 		TestCase: testCase,
 		Results:  toResults(subResponses),
--- a/docker-compose-fail.yml
+++ b/docker-compose-fail.yml
@@ -1,39 +1,65 @@
 crossdock:
-    dns_search: .
     build: .
+    dns_search: .
     links:
-        - alpha
-        - omega
+        - pass
+        - fail1
+        - fail2
+        - fail3
+        - fail4
     environment:
-      - WAIT_FOR=alpha,omega
+      - WAIT_FOR=pass,fail1,fail2,fail3,fail4
 
-      - AXIS_CLIENT=alpha,omega
-      - AXIS_SERVER=alpha,omega
+      - AXIS_CLIENT=pass
+      - AXIS_SERVER=pass
       - AXIS_TRANSPORT=http,tchannel
 
       - BEHAVIOR_DANCE=client,server,transport
       - BEHAVIOR_SING=client,server,transport
 
-      - REPORT=list,json
-      - JSON_REPORT_PATH=/crossdock/report.json
+      - REPORT=list
 
-    volumes:
-      - .crossdock:/crossdock
-
-alpha:
-    dns_search: .
+pass:
     image: breerly/hello-server
+    dns_search: .
     ports:
         - 8080
     environment:
         - HELLO_PORT=8080
         - HELLO_MESSAGE=[{"status":"passed","output":"sup"}]
 
-omega:
-    dns_search: .
+fail1:
     image: breerly/hello-server
+    dns_search: .
     ports:
         - 8080
     environment:
         - HELLO_PORT=8080
-        - HELLO_MESSAGE=[] # no results returned by client
+        - HELLO_MESSAGE="}}!!!invalid json:"
+
+fail2:
+    image: breerly/hello-server
+    dns_search: .
+    ports:
+        - 8080
+    environment:
+        - HELLO_PORT=8080
+        - HELLO_MESSAGE=[{"wrong":"json"}]
+
+fail3:
+    image: breerly/hello-server
+    dns_search: .
+    ports:
+        - 8080
+    environment:
+        - HELLO_PORT=8080
+        - HELLO_MESSAGE=[] # 200, correct json, but no results
+
+fail4:
+    image: breerly/hello-server
+    dns_search: .
+    ports:
+        - 8080
+    environment:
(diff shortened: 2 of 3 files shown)


abhinav:
lgtm

execute/run.go
```diff
@@ -68,6 +68,16 @@ func executeTestCase(testCase plan.TestCase) TestResponse {
...
 
+	if len(subResponses) == 0 {
+		return TestResponse{
+			TestCase: testCase,
+			Results: []Result{{
+				Status: Failed,
+				Output: "client returned 0 results",
+			}},
+		}
+	}
```
abhinav, line 79: A test would be nice

HelloGrayson:
@abhinav I added some additional failure cases in the `docker-compose-fail.yml` and made them run inside `fail.sh`. 

It's not nearly as good as writing better tests in Go, but that's a bigger investment than I'm willing to make right now.

Here's what it looks like if it fails:

```

Expected non-0 exit code for 'explicit failure':
-------------------------------------------------------------------

Crossdock starting...

Configuration:
 - 3 axes:
     client: fail4, pass
     server: pass
  transport: http, tchannel
 - 2 behavior(s):
  dance: client · server · transport
   sing: client · server · transport
 - output: list

Test plan: 8 test(s)

Waiting on WAIT_FOR=[pass fail1 fail2 fail3 fail4]

2016/07/08 08:09:56 HTTP: HEAD http://fail1:8080/
2016/07/08 08:09:56 HTTP: HEAD http://pass:8080/
2016/07/08 08:09:56 HTTP: HEAD http://fail3:8080/
2016/07/08 08:09:56 HTTP: HEAD http://fail2:8080/
2016/07/08 08:09:56 HTTP: HEAD http://fail4:8080/
HTTP: Service fail2:8080 is up
HTTP: Service fail1:8080 is up
HTTP: Service pass:8080 is up
HTTP: Service fail3:8080 is up
HTTP: Service fail4:8080 is up

All services are up after 202.696064ms!

Executing Matrix...

S [dance] fail4→ (server=pass transport=http) ⇒ explicit failure
✓ [dance] pass→ (server=pass transport=http) ⇒ sup
S [dance] fail4→ (server=pass transport=tchannel) ⇒ explicit failure
✓ [dance] pass→ (server=pass transport=tchannel) ⇒ sup
S [sing] fail4→ (server=pass transport=http) ⇒ explicit failure
✓ [sing] pass→ (server=pass transport=http) ⇒ sup
S [sing] fail4→ (server=pass transport=tchannel) ⇒ explicit failure
✓ [sing] pass→ (server=pass transport=tchannel) ⇒ sup

4/4 passed (4/8 skipped)

Tests passed!

-------------------------------------------------------------------
^^^ Expected non-0 exit code for 'explicit failure'

```

HelloGrayson:
@kriskowal could I get some eyes here? 

FYI, this will make yarpc-go fail from now on - we have a couple tests that don't return results at the moment. Want to nip that in the butt.

kriskowal:
lgtm

HelloGrayson:
@abhinav kris and fx didn't seem sold on a client having to return at least 1 result...
what do you think?

bombela:
I don't have a strong opinion really, either works for me!

abhinav:
I could buy that some behaviors with zero failures return nothing. If we go for that, we should alter the output accordingly.

HelloGrayson:
Applying "silence is golden" to a distributed system like this makes me a bit nervous... for example I genuinely wonder if the 3 tests in yarpc-go that return no results are doing anything at all, and there's no find out without digging into the code. 

I think this sensation would get worse if we start doing it everywhere, and I could imagine tests "soft failing" for awhile before we figure it out...

I wonder if we'd instead be doing our customers a favor by forcing always at least 1 assert.

HelloGrayson:
@yurishkuro what's your opinion here - when a behavior runs:
- Client must return at least 1 result? 
- Client can return an `[]`, which will count as `passed`?

execute/run.go
HelloGrayson, line 79: Added :)

yurishkuro:
I am missing a bit of context here - does the client return nothing for a single permutation of behavior axes? Or for the whole behavior execution (all permutations)? On the former I think it should always return P/F/S, and given that, the latter always returns something since it's not possible to have no permutations.

HelloGrayson:
@yurishkuro it's for both. Currently we have some behaviors that return no results (for all permutations), and they are passing anyways. This change is breaking, since those behaviors will start to fail the build once this is merged.

My gut says that this is what we want - clients MUST return at least 1 result for every permutation of every behavior. Returning `[]` from a client will result in the behavior being marked as failed.

I'm going to merge this tomorrow morning if my gut doesn't change :)

HelloGrayson:
I still feel like this is the right call - merging!

PR #77, opened 2016-08-23
OSS Assets

HelloGrayson:
OSS has made the call to transfer HB's logo to Crossdock, adding assets in the repo like we did with YARPC. cc @kriskowal @abhinav


abhinav:
👍

kriskowal:
lgtm

PR #79, opened 2017-01-24
Allow host ready time to be configurable

black-adder:
I'm running into an issue where the hosts are not ready in the default 30 seconds wait time. I'm adding this to allow users to configure the host ready check time.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,7 +23,6 @@ package main
 import (
 	"fmt"
 	"log"
-	"time"
 
 	"github.com/crossdock/crossdock/execute"
 	"github.com/crossdock/crossdock/output"
@@ -50,7 +49,7 @@ func main() {
 	}
 
 	fmt.Printf("\nWaiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
-	execute.Wait(plan.Config.WaitForHosts, time.Duration(30)*time.Second)
+	execute.Wait(plan.Config.WaitForHosts, plan.Config.HostTimeout)
 
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
--- a/main.go
+++ b/main.go
@@ -49,7 +49,7 @@ func main() {
 	}
 
 	fmt.Printf("\nWaiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
-	execute.Wait(plan.Config.WaitForHosts, plan.Config.HostTimeout)
+	execute.Wait(plan.Config.WaitForHosts, plan.Config.WaitForTimeout)
 
 	fmt.Printf("\nExecuting Matrix...\n\n")
 	results := execute.Run(plan)
--- a/plan/config.go
+++ b/plan/config.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	defaultCallTimeout = 5*time.Second
-	defaultHostTimeout = 30*time.Second
+	defaultWaitForTimeout = 30*time.Second
 )
 
 // ReadConfigFromEnviron creates a Config by looking for environment variables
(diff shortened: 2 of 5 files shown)


plan/config.go
```diff
@@ -29,13 +29,17 @@ import (
 	"time"
 )
 
-const defaultCallTimeout = 5
+const (
+	defaultCallTimeout = 5
+	defaultHostTimeout = 30
```
yurishkuro: nit - if you're not using typed values, then indicate the units in the the var name, defaultCallTimeoutInSeconds

yurishkuro (approved):
lgtm

plan/config.go
```diff
@@ -68,7 +76,8 @@ func ReadConfigFromEnviron() (*Config, error) {
 
 	config := &Config{
 		Reports:        reports,
-		CallTimeout:    time.Duration(callTimeout),
+		CallTimeout:    callTimeout*time.Second,
+		HostTimeout:    hostTimeout*time.Second,
```
HelloGrayson: Instead of casting these to time.Second, let's just accept a time.Duration string and use the `time.ParseDuration` function to turn that into a proper `time.Duration`.

This way, the unit will be *in* the configuration:

```
CALL_TIMEOUT=1s
HOST_TIMEOUT=500ms
```

It'll also be much more flexible.

HelloGrayson (changes requested):
Looking really good, let's update to take time duration strings instead of integers, and we should be good to go.

plan/config_test.go
```diff
@@ -33,6 +34,8 @@ func TestReadConfigFromEnviron(t *testing.T) {
 	os.Setenv("AXIS_SERVER", "yarpc-go,yarpc-node")
 	os.Setenv("AXIS_TRANSPORT", "http,tchannel")
 	os.Setenv("BEHAVIOR_ECHO", "client,server,transport")
+	os.Setenv("CALL_TIMEOUT", "10s")
+	os.Setenv("HOST_TIMEOUT", "20s")
```
HelloGrayson: Let's rename this to `WAIT_FOR_TIMEOUT` - this will make it more obvious that it relates to the `WAIT_FOR` setting.

HelloGrayson (changes requested):
Nearly done!

main.go
```diff
@@ -50,7 +49,7 @@ func main() {
 	}
 
 	fmt.Printf("\nWaiting on WAIT_FOR=%v\n\n", plan.Config.WaitForHosts)
-	execute.Wait(plan.Config.WaitForHosts, time.Duration(30)*time.Second)
+	execute.Wait(plan.Config.WaitForHosts, plan.Config.HostTimeout)
```
HelloGrayson: Looks like this needs to be `plan.Config.WaitForTimeout`

black-adder: oops, thanks for the catch

HelloGrayson (approved):
lgtm

HelloGrayson:
lgtm

HelloGrayson:
Aaah, https://lgtm.co/ is deprecated. Give me a bit to remove that integration and then I can merge.

HelloGrayson:
Ok, this is merged. cc @black-adder @yurishkuro 

When the [build](https://travis-ci.org/crossdock/crossdock/builds/195015983) is finished, a new version of Crossdock will be available on [Docker Hub](https://hub.docker.com/r/crossdock/crossdock/tags/).

HelloGrayson:
```
6.34s$ docker push $REPO
The push refers to a repository [docker.io/crossdock/crossdock]
875782e2: digest: sha256:ed63ad800e133fc17031bc102f6e83d34fda44a1ae71ddbad233a8092164ae2d size: 528
latest: digest: sha256:ed63ad800e133fc17031bc102f6e83d34fda44a1ae71ddbad233a8092164ae2d size: 528
travis-454: digest: sha256:ed63ad800e133fc17031bc102f6e83d34fda44a1ae71ddbad233a8092164ae2d size: 528
Done. Your build exited with 0.
```

PR #80, opened 2017-03-27
plan: Fix wrong client axis due to sorting

abhinav:
This fixes a bug where the wrong axis was selected as the client axis
because the list of axes was sorted BEFORE selecting `params[0]` as the
client axis.

CC @breerly @bombela @kriskowal

### DIFF
--- a/plan/config.go
+++ b/plan/config.go
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	defaultCallTimeout = 5*time.Second
-	defaultWaitForTimeout = 30*time.Second
+	defaultCallTimeout    = 5 * time.Second
+	defaultWaitForTimeout = 30 * time.Second
 )
 
 // ReadConfigFromEnviron creates a Config by looking for environment variables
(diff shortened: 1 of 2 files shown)


plan/config.go
```diff
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	defaultCallTimeout = 5*time.Second
-	defaultWaitForTimeout = 30*time.Second
+	defaultCallTimeout    = 5 * time.Second
+	defaultWaitForTimeout = 30 * time.Second
```
abhinav, line 33: this was changed by gofmt :man_shrugging:

kriskowal (approved):

plan/config.go
HelloGrayson: Sure why not :)

plan/config.go
```diff
@@ -92,14 +92,17 @@ func ReadConfigFromEnviron() (*Config, error) {
 func parseBehavior(d string) Behavior {
 	pair := strings.SplitN(d, "=", 2)
 	key := strings.ToLower(pair[0])
+
 	values := strings.Split(pair[1], ",")
 	values = trimCollection(values)
+	clientAxis := values[0]
+	values = values[1:]
 	sort.Strings(values)
```
HelloGrayson, line 100: Really good catch.

plan/config_test.go
```diff
@@ -80,3 +80,39 @@ func TestReadConfigFromEnvironTrimsWhitespace(t *testing.T) {
 
 	assert.Equal(t, config.WaitForHosts, []string{"alpha", "omega"})
 }
+
+func TestParseBehavior(t *testing.T) {
```
HelloGrayson, line 84: Thanks for this test, that should ensure we dont break it in the future.

HelloGrayson (approved):
Thanks, this is a big help.

PR #83, opened 2017-04-12
add support for pruning

### LINKED ISSUE #82: Filtering/pruning of axis per behavior
Currently we take a complete cross product of each axis per
behavior. However we need to filter behavior in scenarios
where we want a particular axis values to not be executed.
Consider the example below:
  AXIS_SERVER=go,java,python,node
  AXIS_CLIENT=go,java,python,node
  AXIS_TRANSPORT=tchannel,http

  BEHAVIOR_RAW=server,client,transport

  Right now, there's no way to skip java for tchannel transport.

Proposed solutions
Solution # 1
  We introduce a new separator that will be used to prune the list.
  BEHAVIOR_RAW=server,client,transport-java

Solution # 2
  We introduce a new behavior along with the behavior that would
represent the axis that you don't want to have.
  AXIS_SERVER_SKIP=java
  AXIS_CLIENT_SKIP=go,java,python,node
  AXIS_TRANSPORT_SKIP=tchannel
  BEHAVIOR_RAW_SKIP=server_skip,client_skip,transport_skip

vivekrsharma:
This fixes https://github.com/crossdock/crossdock/issues/82. Enables using SKIP_ to prune the tests that are not supported.


execute/run.go
```diff
@@ -45,6 +45,16 @@ func Run(plan *plan.Plan) <-chan TestResponse {
 }
 
 func executeTestCase(testCase plan.TestCase) TestResponse {
+	if testCase.Skip {
+		return TestResponse{
+			TestCase: testCase,
+			Results: []Result{{
+				Status: Skipped,
+				// TODO: beefup this error message further?
+				Output: fmt.Sprintf("Skipped due to the filter"),
```
bombela: "Test filtered out"

plan/config.go
```diff
@@ -59,17 +61,25 @@ func ReadConfigFromEnviron() (*Config, error) {
...
+			key, filter := parseSkipBehavior(strings.TrimPrefix(e, behaviorSkipKeyPrefix))
+			filterMap[key] = filter
 		} else if strings.HasPrefix(e, behaviorKeyPrefix) {
 			behavior := parseBehavior(strings.TrimPrefix(e, behaviorKeyPrefix))
 			behaviors = append(behaviors, behavior)
 		}
 	}
 	sort.Sort(axes)
 	sort.Sort(behaviors)
+	behaviors, err := validateAndApplyFilters(filterMap, behaviors)
```
bombela: behaviors.validateAndApplyFilters(filterMap)

plan/config.go
```diff
@@ -59,17 +61,25 @@ func ReadConfigFromEnviron() (*Config, error) {
...
 	var axes Axes
 	var behaviors Behaviors
+	filterMap := make(map[string][]Filter)
 	for _, e := range os.Environ() {
 		if strings.HasPrefix(e, axisKeyPrefix) {
 			axis := parseAxis(strings.TrimPrefix(e, axisKeyPrefix))
 			axes = append(axes, axis)
+		} else if strings.HasPrefix(e, behaviorSkipKeyPrefix) {
+			key, filter := parseSkipBehavior(strings.TrimPrefix(e, behaviorSkipKeyPrefix))
+			filterMap[key] = filter
```
bombela: axes and behaviors are slice, why do you need a map here?

plan/entities.go
```diff
@@ -81,6 +87,7 @@ type TestCase struct {
 	Plan      *Plan
 	Client    string
 	Arguments TestClientArgs
+	Skip      bool
```
bombela: To improve the skip message at the top of the diff, you could add a `SkipReason string`.

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+			return true
+		}
+	}
+	return false
+}
+
+func parseSkipBehavior(d string) (string, []Filter) {
+	pair := strings.SplitN(d, "=", 2)
+	key := strings.ToLower(pair[0])
+	filters := strings.Split(pair[1], ",")
```
bombela: what happens if len(pair) < 2 ?

What about renaming `filters` to `rawFilters` or `strFilters` so to avoid having to name the next variable `filterList`

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+	filters = trimCollection(filters)
```
bombela: you do not need to trim here. you can trim the final axis matches

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+	var filterList []Filter
```
bombela: and then if you have `rawFilters` you can rename this one `filters`, which is what you are trying to get.

Or you could name thems `filtersIn`, `filtersOut`. up to you.

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+	for _, filter := range filters {
```
bombela: same as above this could become:
```go
for _, strfilter := range strFilters {
// or
for _, filterIn := range filtersIn {
```

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+		values := strings.Split(filter, "+")
```
bombela: `values` is more `strMatches` or `matchesIn`

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+		filterItem := Filter{
```
bombela: `filterItem` -> `filter` or `filterOut`

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+					return nil, fmt.Errorf("%v is not defined in axis for %v", match, behavior.Name)
+				}
+			}
+		}
+		behavior.SkipFilters = filters
+	}
+	return behaviors, nil
+}
+
+func axisContainsRule(item string, axis []string) bool {
```
bombela: Rule or Match? Item or axis? axis or axes?

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
 		ClientAxis: clientAxis,
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+func validateAndApplyFilters(filterMap map[string][]Filter, behaviors Behaviors) (Behaviors, error) {
+	for i := range behaviors {
```
bombela: `for _, behavior := range behaviors {`

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
 		ClientAxis: clientAxis,
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+func validateAndApplyFilters(filterMap map[string][]Filter, behaviors Behaviors) (Behaviors, error) {
```
bombela: `filterMap` could be renamed `filtersByBehavior`

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+	for i := range behaviors {
+		behavior := &behaviors[i]
+		filters := filterMap[behavior.Name]
+		for _, filter := range filters {
+			for match := range filter.AxisMatches {
```
bombela: `match` is really `axisToMatch` or just `axis`.

plan/config.go
```diff
@@ -104,10 +114,54 @@ func parseBehavior(d string) Behavior {
...
+				if match != behavior.ClientAxis && !axisContainsRule(match, behavior.ParamsAxes) {
```
bombela: `if !behavior.HasAxis(axis) { err }` 

I believe you can then replace axisContainsRule by this new HasAxis method.

plan/new.go
```diff
@@ -52,7 +52,23 @@ func buildTestCases(plan *Plan) []TestCase {
...
+			for _, filter := range behavior.SkipFilters {
+				// Each filter should be a complete match
+				isCompleteMatch := true
+				for key, value := range filter.AxisMatches {
+					if (key == behavior.ClientAxis && value != client) ||
+						testArgs[key] != value {
+						isCompleteMatch = false
+					}
+				}
+				if isCompleteMatch {
```
bombela: no need to continue lopping after a complete match.

plan/new.go
```diff
@@ -52,7 +52,23 @@ func buildTestCases(plan *Plan) []TestCase {
...
 				Client:    client,
 				Arguments: testArgs,
+				Skip:      false,
 			}
+
+			for _, filter := range behavior.SkipFilters {
+				// Each filter should be a complete match
+				isCompleteMatch := true
+				for key, value := range filter.AxisMatches {
+					if (key == behavior.ClientAxis && value != client) ||
```
bombela: no need to continue looping if a match fails.

plan/new.go
```diff
@@ -52,7 +52,23 @@ func buildTestCases(plan *Plan) []TestCase {
 				Plan:      plan,
 				Client:    client,
 				Arguments: testArgs,
+				Skip:      false,
 			}
+
+			for _, filter := range behavior.SkipFilters {
+				// Each filter should be a complete match
+				isCompleteMatch := true
```
bombela: A cool way to represent matches in a list like that is to use a counter. `matchCnt := 0`. Increment it at every successful match. After, if `matchCnt == len(filter.AxisMatches)`, you know you got a full match. It has the nice effect if being really easy to read and reason about. But its up to you and mostly a matter of personal taste.

plan/new.go
bombela: you are only testing the client axis here, not all axes.

Also filter.AxisMatches is a map, but you don't seem to use any properties of the map either.

bombela (changes requested):

plan/config.go
vivekrsharma: Okay if you look down below, we need to get the slice of filters per behavior when we iterate over behavior. The lookup becomes easier with map rather than having to do O( size of behavior * size of filter).

plan/config.go
vivekrsharma: we update behaviors as well right?

plan/config.go
vivekrsharma: Yep, I'd a good learning here. If I do that, the object of behavior that is updated is not reflected in calling function.

plan/config.go
vivekrsharma: I like axisToMatch better.

plan/config.go
vivekrsharma: Axis - Singular, Axes - plural and updated the function accordingly. Also this function now becomes part of the entities.go ( on behavior)

plan/config.go
vivekrsharma: Good point. I also saw we behaved in similar way up while parsing axis and behaviors. 

Yes, the renamed helped with readability. Thanks.

plan/config.go
vivekrsharma: I'm sticking with rawFilters and filters. It is not an overkill and serves readability.

plan/new.go
vivekrsharma: Good catch

plan/new.go
vivekrsharma: Nope. I'm checking all the axis. testArgs does that. testArgs has everything but 0th argument that is client axis.

Well, I'm using the map property in validateAndApplyFilters when I need to get all the skip filters for a particular behavior.
Note the map is behavior -> slice of Filters. With each filter in Filters being a map.

plan/entities.go
vivekrsharma: Yep good idea. Also added a naive message saying filtered out due to pruning. We can beef it up if needed.

plan/new.go
vivekrsharma: I like the readability more. Changed it to that. However note that this tries to match every key value in the map and is less efficient.

glide.lock
```diff
@@ -1,19 +1,36 @@
...
+updated: 2017-04-12T11:40:29.878866333-07:00
 imports:
 - name: github.com/fatih/color
   version: 533cd7fd8a85905f67a1753afb4deddc85ea174f
+- name: github.com/mattn/go-colorable
+  version: 9cbef7c35391cca05f15f8181dc0b18bc9736dbb
+  repo: https://github.com/mattn/go-colorable
+- name: github.com/mattn/go-isatty
+  version: 56b76bdf51f7708750eac80fa38b952bb9f32639
+  repo: https://github.com/mattn/go-isatty
```
vivekrsharma: This and file below would go away once PR # 84 is merged in.

plan/config.go
```diff
@@ -36,13 +36,14 @@ const (
...
-		axisKeyPrefix     = "AXIS_"
-		behaviorKeyPrefix = "BEHAVIOR_"
-		jsonReportPathKey = "JSON_REPORT_PATH"
+		reportKey             = "REPORT"
+		callTimeoutKey        = "CALL_TIMEOUT"
+		waitForTimeoutKey     = "WAIT_FOR_TIMEOUT"
+		waitKey               = "WAIT_FOR"
+		axisKeyPrefix         = "AXIS_"
+		behaviorKeyPrefix     = "BEHAVIOR_"
+		behaviorSkipKeyPrefix = "BEHAVIOR_SKIP_"
```
kriskowal: Please make this just `"SKIP_"`. Building on the `"BEHAVIOR_"` prefix has the unintended consequence of subtracting from the range of expressible behavior names.

plan/config.go
```diff
@@ -104,10 +113,29 @@ func parseBehavior(d string) Behavior {
 		ClientAxis: clientAxis,
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+func parseSkipBehavior(d string) (string, []Filter) {
```
kriskowal: Please add a comment describing the grammar and semantics for the skip string here or link elsewhere.

plan/config_test.go
```diff
@@ -116,3 +126,46 @@ func TestParseBehavior(t *testing.T) {
...
 }
+
+func TestParseSkipBehavior(t *testing.T) {
+	tests := []struct {
+		give       string
+		wantFilter []Filter
+		wantKey    string
+	}{
+		{
+			give: "foo=client=c+server=b",
```
kriskowal: perhaps give descriptions for each of these table test cases (and use `t.Run(tt.desc, func(t) {})`), e.g., "skip all cases in the 'foo' behavior where the CLIENT is 'foo' and SERVER is 'b') (did I get that right?)

docker-compose.yml
```diff
@@ -13,6 +13,8 @@ crossdock:
 
       - BEHAVIOR_DANCE=client,server,transport
       - BEHAVIOR_SING=client,server,transport
+      - BEHAVIOR_RUN=client,server,transport
+      - BEHAVIOR_SKIP_RUN=client=alpha+server=alpha
```
kriskowal: This might be more legible with colons and commas. We already use comma as a delimiter, and `=` for the assignment.

```
SKIP_RUN=client:alpha,server:alpha
```

execute/run.go
```diff
@@ -45,6 +45,15 @@ func Run(plan *plan.Plan) <-chan TestResponse {
 }
 
 func executeTestCase(testCase plan.TestCase) TestResponse {
+	if testCase.Skip {
+		return TestResponse{
+			TestCase: testCase,
+			Results: []Result{{
+				Status: Skipped,
+				Output: fmt.Sprintf("Test filtered out"),
```
kriskowal: Or merely, `"Skipped %s"` for case.

plan/new.go
```diff
@@ -52,7 +52,25 @@ func buildTestCases(plan *Plan) []TestCase {
...
+				matchCnt := 0
+				for key, value := range filter.AxisMatches {
+					if (key == behavior.ClientAxis && value == client) ||
+						testArgs[key] == value {
+						matchCnt++
+					}
+				}
+				if matchCnt == len(filter.AxisMatches) {
+					t.Skip = true
+					t.SkipReason = "Filtered out in pruning"
```
kriskowal: Or merely `"Pruned by SKIP_%s"` rule.

plan/new_test.go
```diff
@@ -180,3 +180,142 @@ func TestNew(t *testing.T) {
...
+			{Name: "client", Values: []string{"alpha", "omega"}},
+			{Name: "server", Values: []string{"alpha", "omega"}},
+			{Name: "transport", Values: []string{"http", "tchannel"}},
+		},
+		Behaviors: []Behavior{
+			{
+				Name:       "dance",
+				ClientAxis: "client",
+				ParamsAxes: []string{"server", "transport"},
+				SkipFilters: []Filter{
```
kriskowal: SkipFilter is redundant. Consider just "Filters".

docker-compose.yml
kriskowal: Perhaps we should call this `FILTER_` or `PRUNE_` to distinguish from tests skipped for other reasons.

kriskowal (changes requested):
Logic looks good. Requesting some æsthetic changes to make the usage more clear.

docker-compose.yml
vivekrsharma: Actually, we wanted to maintain same word "skip" and provide additional reason to why we were skipping. I'd like to stick to SKIP_ but I'm flexible if others feel the other way.

plan/config.go
vivekrsharma: SKIP_ looks good to me.

docker-compose.yml
vivekrsharma: Yes, so : is used for assignment and + is used to separated multiple axis conditions in a filter.
Multiple filters are separated by a comma.

plan/config_test.go
vivekrsharma: Right on. Yes, that would make it more readable.

execute/run.go
vivekrsharma: Nice suggestions on readability.

plan/new.go
vivekrsharma: Nice suggestions on readability.

plan/config.go
```diff
@@ -59,17 +60,25 @@ func ReadConfigFromEnviron() (*Config, error) {
...
 	var axes Axes
 	var behaviors Behaviors
+	filterMap := make(map[string][]Filter)
 	for _, e := range os.Environ() {
 		if strings.HasPrefix(e, axisKeyPrefix) {
 			axis := parseAxis(strings.TrimPrefix(e, axisKeyPrefix))
 			axes = append(axes, axis)
+		} else if strings.HasPrefix(e, skipKeyPrefix) {
+			key, filter := parseSkipBehavior(strings.TrimPrefix(e, skipKeyPrefix))
+			filterMap[key] = filter
```
abhinav: optional: Often when you have a bunch of `if-else-if` blocks, it's cleaner to
use a `switch` statement without a value:

```
switch {
case strings.HasPrefix(e, axisKeyPrefix):
  ...
case strings.HasPrefix(e, skipKeyPrefix):
  ...
case strings.HasPrefix(e, behaviorKeyPrefix):
  ...
}
```

plan/config.go
```diff
@@ -59,17 +60,25 @@ func ReadConfigFromEnviron() (*Config, error) {
...
 		} else if strings.HasPrefix(e, behaviorKeyPrefix) {
 			behavior := parseBehavior(strings.TrimPrefix(e, behaviorKeyPrefix))
 			behaviors = append(behaviors, behavior)
 		}
 	}
 	sort.Sort(axes)
 	sort.Sort(behaviors)
+	err := behaviors.validateAndApplyFilters(filterMap)
+	if err != nil {
```
abhinav: Inline the `err` into the if statement. Usually, if you can inline it, it's usually a good idea to do it.

```
if err := behaviors.validateAndApplyFilters(..); err != nil {
  return nil, err
}
```

plan/config.go
```diff
@@ -104,10 +113,32 @@ func parseBehavior(d string) Behavior {
 		ClientAxis: clientAxis,
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+// Skip string is of format SKIP_RUN=axis1:value1+axis2:value2,axis1:value3 where
+// SKIP_ is followed by behavior name and multiple filters are separated by comma.
+// Each filter is a logical AND and complete match is supported.
```
abhinav: Just to clarify, do you mean each set of `+`s is a logical AND and the `,` separate filters which are ORed?

If I do,

    server:java+transport:tchannel,client:java+transport:tchannel

My expectation would be that the behavior will be skipped if,

    (server == java && transport == tchannel) || (client == java && transport == tchannel)

If this is accurate, the comment is outdated or hard-to-understand.

plan/config.go
```diff
@@ -104,10 +113,32 @@ func parseBehavior(d string) Behavior {
...
+func parseSkipBehavior(d string) (string, []Filter) {
+	pair := strings.SplitN(d, "=", 2)
+	key := strings.ToLower(pair[0])
+	rawFilters := strings.Split(pair[1], ",")
+	var filters []Filter
```
abhinav: You know the number of filters in advance. Use `make([]Filter, 0, len(rawFilters))` to allocate a slice with just enough capacity.

plan/config.go
```diff
@@ -104,10 +113,32 @@ func parseBehavior(d string) Behavior {
...
+	for _, rawFilter := range rawFilters {
+		strMatches := strings.Split(rawFilter, "+")
+		filter := Filter{
+			AxisMatches: map[string]string{},
```
abhinav: Nit: Prefer `make(map[string]string)` over the `map[string]string{}` form. The latter form is usually used if you have values to put into the map right off the bat.

plan/config.go
```diff
@@ -104,10 +113,32 @@ func parseBehavior(d string) Behavior {
...
+		}
+		for _, strMatch := range strMatches {
+			tuple := strings.Split(strMatch, ":")
```
abhinav: Use SplitN with 2. You want to ensure that `foo:bar:baz` isn't interpreted as `foo == bar` instead of `foo == bar:baz`.

(Also, add a test for this case.)

plan/config.go
```diff
@@ -104,10 +113,32 @@ func parseBehavior(d string) Behavior {
...
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+// Skip string is of format SKIP_RUN=axis1:value1+axis2:value2,axis1:value3 where
+// SKIP_ is followed by behavior name and multiple filters are separated by comma.
+// Each filter is a logical AND and complete match is supported.
+func parseSkipBehavior(d string) (string, []Filter) {
```
abhinav: This function is quite fragile. I can make it panic with bad input: `parseSkipBehavior("FOO")`, `parseSkipBehavior("FOO=bar+baz")` at the least. The number given to SplitN determines the maximum number of values returned, not the exact. You can receive fewer values. Don't index into a slice without verifying its length first.

You'll want to make this function return an `error` in case of badly formed inputs.

plan/config_test.go
```diff
@@ -116,3 +126,49 @@ func TestParseBehavior(t *testing.T) {
...
+func TestParseSkipBehavior(t *testing.T) {
+	tests := []struct {
+		give       string
+		desc       string
+		wantFilter []Filter
+		wantKey    string
+	}{
+		{
+			give: "foo=client:c+server:b",
+			desc: "skip all cases in the 'foo' behavior and where client axis  has value 'c' and server axis has value 'b'.",
```
abhinav: Test descriptions don't need to be this verbose; they especially don't need to describe all the inputs of the test case. Just say what kind of scenario this is testing. "single filter" in this case, "multiple filters" for the next one.

plan/config_test.go
```diff
@@ -116,3 +126,49 @@ func TestParseBehavior(t *testing.T) {
...
+				},
+				{
+					AxisMatches: map[string]string{
+						"c": "d",
+					},
+				},
+			},
+			wantKey: "x",
+		},
+	}
```
abhinav: Add tests cases for empty strings and empty filters (like `x=a:b,,c:d`).

plan/entities.go
```diff
@@ -55,11 +58,31 @@ func (a Axes) Index() map[string]Axis {
 	return axes
 }
 
+// Filter is collection of axis to escape
```
abhinav: - Plural of axis is axes.
- End statements in docs with periods. (I'm aware that some of the existing code isn't doing that and we can fix that later.)
- This comment doesn't quite tell me what a filter is. What's "axis to escape"?

plan/entities.go
```diff
@@ -55,11 +58,31 @@ func (a Axes) Index() map[string]Axis {
...
+type Filter struct {
+	AxisMatches map[string]string
```
abhinav: Explain what this attribute contains. What's it a mapping of? How does it affect behavior?

plan/entities.go
```diff
@@ -55,11 +58,31 @@ func (a Axes) Index() map[string]Axis {
...
+
 // Behavior represents the test behavior that will be triggered by crossdock
 type Behavior struct {
 	Name       string
 	ClientAxis string
 	ParamsAxes []string
+	Filters    []Filter
+}
+
+// HasAxis checks and returns true if passes axis is part of behavior, false otherwise.
```
abhinav: > [..] if the *passed* axis is part of *the* behavior

plan/entities.go
```diff
@@ -55,11 +58,31 @@ func (a Axes) Index() map[string]Axis {
...
+func (b Behavior) HasAxis(axis string) bool {
+	return axis == b.ClientAxis || axesContainsAxis(axis, b.ParamsAxes)
+}
+
+func axesContainsAxis(axisToFind string, axes []string) bool {
```
abhinav: Honestly, I wouldn't make this a function. Just stick this in HasAxis for now.

plan/entities.go
```diff
@@ -68,6 +91,21 @@ type Behaviors []Behavior
 func (b Behaviors) Len() int           { return len(b) }
 func (b Behaviors) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
 func (b Behaviors) Less(i, j int) bool { return b[i].Name < b[j].Name }
+func (b Behaviors) validateAndApplyFilters(filterByBehavior map[string][]Filter) error {
```
abhinav: Add empty line before the method. The previous three are clumped together because they're busy-work to implement the `sort.Interface` interface.

plan/config.go
```diff
@@ -59,17 +60,25 @@ func ReadConfigFromEnviron() (*Config, error) {
 
 	var axes Axes
 	var behaviors Behaviors
+	filterMap := make(map[string][]Filter)
 	for _, e := range os.Environ() {
 		if strings.HasPrefix(e, axisKeyPrefix) {
 			axis := parseAxis(strings.TrimPrefix(e, axisKeyPrefix))
 			axes = append(axes, axis)
+		} else if strings.HasPrefix(e, skipKeyPrefix) {
+			key, filter := parseSkipBehavior(strings.TrimPrefix(e, skipKeyPrefix))
```
abhinav: nit: s/filter/filters

parseSkipBehavior is returning multiple filters

plan/entities.go
```diff
@@ -68,6 +91,21 @@ type Behaviors []Behavior
...
+	for i := range b {
+		behavior := &b[i]
```
abhinav: Use the two-result form of `range`

```
for i, behavior := range b {
  ...

  behavior.Filters = filters
  b[i] = behavior
}
```

plan/new.go
```diff
@@ -52,7 +57,25 @@ func buildTestCases(plan *Plan) []TestCase {
...
+				matchCnt := 0
+				for key, value := range filter.AxisMatches {
+					if (key == behavior.ClientAxis && value == client) ||
+						testArgs[key] == value {
+						matchCnt++
+					}
+				}
+				if matchCnt == len(filter.AxisMatches) {
+					t.Skip = true
+					t.SkipReason = fmt.Sprintf("Pruned by SKIP_%s", strings.ToUpper(behavior.Name))
```
abhinav: IMO, you should mention the exact filter that caused this to be skipped.

abhinav (changes requested):

plan/config.go
vivekrsharma: Updated the comment and moved it to README.md as discussed offline.

plan/new.go
vivekrsharma: Done.

README.md
```diff
@@ -71,6 +71,22 @@ Beginning matrix of tests...
   PASSED | omega  | ok       | slow  | run
 ```
 
+## Support for filtering using SKIP_
+
+Skip clauses are comprised of zero or more filter expressions separated by commas. These are or-ed together.
+```
+SKIP_RUN=filter1,filter2,...,filterN
+```
+Where each filter consists or zero or more axis matchers separated by ``` + ``` s. These are and-ed together.
```
bombela: A single ` is enough for inlined \<pre\>.

bombela (approved):

plan/config.go
```diff
@@ -59,17 +61,31 @@ func ReadConfigFromEnviron() (*Config, error) {
...
+		switch {
+		case strings.HasPrefix(e, axisKeyPrefix):
 			axis := parseAxis(strings.TrimPrefix(e, axisKeyPrefix))
 			axes = append(axes, axis)
-		} else if strings.HasPrefix(e, behaviorKeyPrefix) {
+			break
+		case strings.HasPrefix(e, skipKeyPrefix):
+			key, filters, err := parseSkipBehavior(strings.TrimPrefix(e, skipKeyPrefix))
+			if err != nil {
+				return nil, err
```
abhinav: More information in the error message please.

    return nil, fmt.Errorf("failed to parse filters from %q: %v", e, err)

plan/config.go
```diff
@@ -59,17 +61,31 @@ func ReadConfigFromEnviron() (*Config, error) {
...
 	var behaviors Behaviors
+	filterMap := make(map[string][]Filter)
 	for _, e := range os.Environ() {
-		if strings.HasPrefix(e, axisKeyPrefix) {
+		switch {
+		case strings.HasPrefix(e, axisKeyPrefix):
 			axis := parseAxis(strings.TrimPrefix(e, axisKeyPrefix))
 			axes = append(axes, axis)
-		} else if strings.HasPrefix(e, behaviorKeyPrefix) {
+			break
```
abhinav: In Go switch statements, `break` is the default. You don't need to add these.

Your linter or `go vet` should have complained about this.

plan/config.go
```diff
@@ -104,10 +120,35 @@ func parseBehavior(d string) Behavior {
...
 		ParamsAxes: values,
 	}
-
 	return behavior
 }
 
+func parseSkipBehavior(d string) (string, []Filter, error) {
+	pair := strings.SplitN(d, "=", 2)
+	if len(pair) != 2 {
+		return "", nil, fmt.Errorf("Invalid input: %s", d)
```
abhinav: Error messages should start with lower case.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

abhinav: Also, when referencing an input string in error messages, it's often useful to use `%q` rather than `%s` because it will quote the string so you'll have a reasonable error message even if the string is empty.

    invalid input ""

abhinav: Further, 'invalid input' isn't a useful error message. Say something like,

   missing '=' in environment variable

abhinav (commented):
@bombela Is going to take another look first.

plan/config.go
```diff
@@ -59,17 +61,31 @@ func ReadConfigFromEnviron() (*Config, error) {
...
+		case strings.HasPrefix(e, behaviorKeyPrefix):
 			behavior := parseBehavior(strings.TrimPrefix(e, behaviorKeyPrefix))
 			behaviors = append(behaviors, behavior)
+			break
 		}
 	}
 	sort.Sort(axes)
 	sort.Sort(behaviors)
+	if err := behaviors.validateAndApplyFilters(filterMap); err != nil {
+		return nil, err
```
bombela: `return nil, fmt.Errorf("failed to validate filters: %v", err)`

plan/config.go
```diff
@@ -104,10 +120,35 @@ func parseBehavior(d string) Behavior {
...
-
 	return behavior
 }
 
+func parseSkipBehavior(d string) (string, []Filter, error) {
+	pair := strings.SplitN(d, "=", 2)
+	if len(pair) != 2 {
+		return "", nil, fmt.Errorf("Invalid input: %s", d)
+	}
+	key := strings.ToLower(pair[0])
```
bombela: rename `key` to `behaviorName`

plan/config.go
```diff
@@ -104,10 +120,35 @@ func parseBehavior(d string) Behavior {
...
+		for _, rawMatch := range rawMatches {
+			tuple := strings.SplitN(rawMatch, ":", 2)
+			if len(tuple) != 2 {
+				return "", nil, fmt.Errorf("Invalid filter definition: %s in input %s", rawMatch, d)
+			}
+			tuple = trimCollection(tuple)
+			filter.AxisMatches[tuple[0]] = tuple[1]
+		}
+		filters = append(filters, filter)
+	}
```
bombela: Maybe you should return an error here, if `len(filters)==0` since it would be weird to have a SKIP behavior without any filters.

plan/config.go
```diff
@@ -104,10 +120,35 @@ func parseBehavior(d string) Behavior {
...
+			AxisMatches: make(map[string]string),
+		}
+		for _, rawMatch := range rawMatches {
+			tuple := strings.SplitN(rawMatch, ":", 2)
+			if len(tuple) != 2 {
+				return "", nil, fmt.Errorf("Invalid filter definition: %s in input %s", rawMatch, d)
+			}
+			tuple = trimCollection(tuple)
+			filter.AxisMatches[tuple[0]] = tuple[1]
+		}
```
bombela: return an error if `len(filter.AxisMatches) == 0` (use %q on rawFilter + even the index for a really nice error msg)

plan/config.go
```diff
@@ -104,10 +120,35 @@ func parseBehavior(d string) Behavior {
...
+		filter := Filter{
+			AxisMatches: make(map[string]string),
+		}
+		for _, rawMatch := range rawMatches {
+			tuple := strings.SplitN(rawMatch, ":", 2)
+			if len(tuple) != 2 {
+				return "", nil, fmt.Errorf("Invalid filter definition: %s in input %s", rawMatch, d)
+			}
+			tuple = trimCollection(tuple)
+			filter.AxisMatches[tuple[0]] = tuple[1]
```
bombela: check before this, if tuple[*] are not empty.

plan/config_test.go
```diff
@@ -116,3 +127,74 @@ func TestParseBehavior(t *testing.T) {
 		assert.Equal(t, tt.want, parseBehavior(tt.give))
 	}
 }
+
+func TestParseSkipBehavior(t *testing.T) {
+	tests := []struct {
+		give       string
+		desc       string
+		wantFilter []Filter
+		wantKey    string
```
bombela: what is wantKey?

plan/config_test.go
```diff
@@ -116,3 +127,74 @@ func TestParseBehavior(t *testing.T) {
...
+			wantKey:   "foo",
+			wantError: nil,
+		},
+		{
+			give:       "x=a:b,,c:d",
+			desc:       "invalid filters",
+			wantFilter: nil,
+			wantKey:    "",
+			wantError:  fmt.Errorf("Invalid filter definition:  in input x=a:b,,c:d"),
+		},
```
bombela: add an extra test with spaces all over the place.

plan/config_test.go
```diff
@@ -116,3 +127,74 @@ func TestParseBehavior(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		key, filter, err := parseSkipBehavior(tt.give)
```
bombela: filter plurial

bombela: handle err

plan/config_test.go
```diff
@@ -116,3 +127,74 @@ func TestParseBehavior(t *testing.T) {
 		assert.Equal(t, tt.want, parseBehavior(tt.give))
 	}
 }
+
+func TestParseSkipBehavior(t *testing.T) {
+	tests := []struct {
+		give       string
+		desc       string
+		wantFilter []Filter
```
bombela: wantFilter plurial

plan/entities.go
```diff
@@ -55,11 +58,32 @@ func (a Axes) Index() map[string]Axis {
 	return axes
 }
 
+// Filter specifies criteria for skipping specific test cases of a behavior.
+// AxisMatches holds a mapping of behavior parameters to values.
```
bombela: AxisMatches holds a mapping axis name to axis value.

plan/entities.go
```diff
@@ -55,11 +58,32 @@ func (a Axes) Index() map[string]Axis {
...
+
 // Behavior represents the test behavior that will be triggered by crossdock
 type Behavior struct {
 	Name       string
 	ClientAxis string
 	ParamsAxes []string
+	Filters    []Filter
+}
+
+// HasAxis checks and returns true if the passed axis is part of behavior, false otherwise.
```
bombela: is part of behavior -> is referenced by the behavior

plan/entities.go
```diff
@@ -69,6 +93,22 @@ func (b Behaviors) Len() int           { return len(b) }
 func (b Behaviors) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
 func (b Behaviors) Less(i, j int) bool { return b[i].Name < b[j].Name }
 
+func (b Behaviors) validateAndApplyFilters(filtersByBehavior map[string][]Filter) error {
+	for i, behavior := range b {
+		filters := filtersByBehavior[behavior.Name]
+		for _, filter := range filters {
+			for axisToMatch := range filter.AxisMatches {
```
bombela: axisToMatch -> axisName

plan/new.go
```diff
@@ -52,7 +58,34 @@ func buildTestCases(plan *Plan) []TestCase {
 				Plan:      plan,
 				Client:    client,
 				Arguments: testArgs,
+				Skip:      false,
+			}
+
+			for _, filter := range behavior.Filters {
+				// Each filter should be a complete match
+				matchCnt := 0
+				for key, value := range filter.AxisMatches {
```
bombela: key -> axisName
value -> axisValue

bombela: AxisMatches really doesn't need to be a map since you never use any of its [data structure] properties.

AxisMatches should really be: matched: []AxisMatches and type AxisMatches struct { AxisName string, AxisValue string } something like that, of course adjust.

plan/new.go
```diff
@@ -52,7 +58,34 @@ func buildTestCases(plan *Plan) []TestCase {
...
+					if (key == behavior.ClientAxis && value == client) ||
+						testArgs[key] == value {
+						matchCnt++
+					}
+				}
+				if matchCnt == len(filter.AxisMatches) {
+					t.Skip = true
+					var formattedMatches []string
+					var keys []string
```
bombela: keys is too generic of a name

plan/new.go
```diff
@@ -52,7 +58,34 @@ func buildTestCases(plan *Plan) []TestCase {
...
+					for k := range filter.AxisMatches {
+						keys = append(keys, k)
+					}
+					sort.Strings(keys)
```
bombela: why do you sort the axes names here?

bombela (changes requested):

plan/entities.go
vivekrsharma: done.

plan/new.go
vivekrsharma: Done.

plan/config.go
vivekrsharma: As discussed in person, this would never happen because of way we're handling errors.

plan/config.go
vivekrsharma: As discussed in person, this would never happen because of way we're handling errors.

plan/config_test.go
vivekrsharma: wantBehaviorName. Fixed.

plan/config_test.go
vivekrsharma: Did you mean comparing it with expected error?

bombela: You must check `err`. For example: `if assert.NoError(t, err) { further assertions }`

vivekrsharma: Thanks for detailed offline explanation. Fixed as discussed.

plan/config.go
```diff
@@ -164,15 +164,16 @@ func parseAxis(d string) Axis {
 	return axis
 }
 
+// TODO: Return an error if input is malformed.
 func validateConfig(config *Config) error {
 	axes := config.Axes.Index()
 	for _, behavior := range config.Behaviors {
 		if _, ok := axes[behavior.ClientAxis]; !ok {
-			return errors.New("Can't find AXIS environment for: " + behavior.ClientAxis)
+			return errors.New("can't find AXIS environment for: " + behavior.ClientAxis)
```
bombela: fmt.Errorf

plan/config.go
```diff
@@ -164,15 +164,16 @@ func parseAxis(d string) Axis {
...
 		}
 		for _, param := range behavior.ParamsAxes {
 			if _, ok := axes[param]; !ok {
-				return errors.New("Can't find AXIS environment for: " + param)
+				return errors.New("can't find AXIS environment for: " + param)
```
bombela: fmt.Errorf

plan/entities.go
```diff
@@ -59,10 +59,16 @@ func (a Axes) Index() map[string]Axis {
 }
 
 // Filter specifies criteria for skipping specific test cases of a behavior.
-// AxisMatches holds a mapping of behavior parameters to values.
+// AxisMatches holds a mapping axis name to axis value.
```
bombela: No need to document AxisMatches. AxisMatch is documented below anyway.

plan/entities.go
```diff
@@ -59,10 +59,16 @@ func (a Axes) Index() map[string]Axis {
...
 // All test cases for a behavior where all parameter values match the map will be skipped.
```
bombela: This comment should be updated.

bombela (changes requested):

plan/config.go
```diff
@@ -169,11 +170,11 @@ func validateConfig(config *Config) error {
 	axes := config.Axes.Index()
 	for _, behavior := range config.Behaviors {
 		if _, ok := axes[behavior.ClientAxis]; !ok {
-			return errors.New("can't find AXIS environment for: " + behavior.ClientAxis)
+			return fmt.Errorf("can't find AXIS environment for: " + behavior.ClientAxis)
```
bombela: ` fmt.Errorf("can't find AXIS environment for: %s", behavior.ClientAxis)`

plan/config.go
```diff
@@ -169,11 +170,11 @@ func validateConfig(config *Config) error {
...
 		}
 		for _, param := range behavior.ParamsAxes {
 			if _, ok := axes[param]; !ok {
-				return errors.New("can't find AXIS environment for: " + param)
+				return fmt.Errorf("can't find AXIS environment for: " + param)
```
bombela: same as above

plan/config_test.go
```diff
@@ -221,25 +221,46 @@ func TestParseSkipBehavior(t *testing.T) {
...
 		{
 			give:             "x= :",
 			desc:             "invalid filters",
 			wantFilters:      nil,
 			wantBehaviorName: "",
-			wantError:        fmt.Errorf("match \" :\" in input x= : is not of form 'key:value'"),
+			wantError:        fmt.Errorf("match \" :\" in input \"x= :\" is empty"),
+		},
+		{
+			give:             "x=     :",
```
bombela: ` "x=     :  "`

bombela (changes requested):

plan/config.go
```diff
@@ -170,11 +170,11 @@ func validateConfig(config *Config) error {
 	axes := config.Axes.Index()
 	for _, behavior := range config.Behaviors {
 		if _, ok := axes[behavior.ClientAxis]; !ok {
-			return fmt.Errorf("can't find AXIS environment for: " + behavior.ClientAxis)
+			return fmt.Errorf("can't find AXIS environment for: %s" + behavior.ClientAxis)
```
bombela: `fmt.Errorf("can't find AXIS environment for: %s", behavior.ClientAxis)`

plan/config.go
```diff
@@ -170,11 +170,11 @@ func validateConfig(config *Config) error {
...
 		}
 		for _, param := range behavior.ParamsAxes {
 			if _, ok := axes[param]; !ok {
-				return fmt.Errorf("can't find AXIS environment for: " + param)
+				return fmt.Errorf("can't find AXIS environment for: %s" + param)
```
bombela: same as above

bombela (approved):

plan/config.go
```diff
@@ -104,10 +117,39 @@ func parseBehavior(d string) Behavior {
...
+	filters := make([]Filter, 0, len(rawFilters))
+	for _, rawFilter := range rawFilters {
+		rawMatches := strings.Split(rawFilter, "+")
+		filter := Filter{
+			AxisMatches: make([]AxisMatch, 0, len(rawMatches)),
+		}
+		for _, rawMatch := range rawMatches {
+			tuple := strings.SplitN(rawMatch, ":", 2)
+			if len(tuple) != 2 {
+				return "", nil, fmt.Errorf("match %q in input %q is not of form 'key:value'", rawMatch, d)
```
abhinav: s/match/matcher

plan/config.go
```diff
@@ -104,10 +117,39 @@ func parseBehavior(d string) Behavior {
...
+			}
+			axisName := strings.TrimSpace(tuple[0])
+			axisValue := strings.TrimSpace(tuple[1])
+			if axisName == "" || axisValue == "" {
+				return "", nil, fmt.Errorf("match %q in input %q is empty", rawMatch, d)
```
abhinav: > invalid matcher "%q": axis name and value are required

plan/config.go
```diff
@@ -123,15 +165,16 @@ func parseAxis(d string) Axis {
 	return axis
 }
 
+// TODO: Return an error if input is malformed.
 func validateConfig(config *Config) error {
 	axes := config.Axes.Index()
 	for _, behavior := range config.Behaviors {
 		if _, ok := axes[behavior.ClientAxis]; !ok {
-			return errors.New("Can't find AXIS environment for: " + behavior.ClientAxis)
+			return fmt.Errorf("can't find AXIS environment for: %s", behavior.ClientAxis)
```
abhinav, line 173: thanks for fixing this

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			wantBehaviorName: "x",
+			wantError:        nil,
+		},
+		{
+			give: "foo=client:weird:name+server:b",
+			desc: "single filter testing name separation.",
```
abhinav: nit: no need to add `.` to test case descriptions

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+						},
+						AxisMatch{
+							Name:  "server",
+							Value: "b",
+						},
+					},
+				},
+			},
+			wantBehaviorName: "foo",
+			wantError:        nil,
```
abhinav: nit: sounds like the test cases can have either `wantFilters` and `wantBehaviorName` set or just `wantError`. It's okay to leave the parts that you're not testing (`wantError` in this case) to the zero value by not mentioning it in the struct literal.

```
{
  desc: ..,
  give: ..,
  wantFilters: ..,
  wantBehaviorName: ..,
}
```

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+	tests := []struct {
+		give             string
+		desc             string
+		wantFilters      []Filter
+		wantBehaviorName string
+		wantError        error
+	}{
+		{
+			give: "foo=client:c+server:b",
+			desc: "single filter.",
```
abhinav: nit: 'desc' should usually be the first one in the list because it's the name of the test case

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			wantBehaviorName: "foo",
+			wantError:        nil,
+		},
+		{
+			give:             "x=a:b,,c:d",
+			desc:             "invalid filters",
```
abhinav: nit:

> invalid: empty filter

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+		{
+			give:             "  x    ",
+			desc:             "invalid filters",
+			wantFilters:      nil,
+			wantBehaviorName: "",
+			wantError:        fmt.Errorf("missing '=' in the input: \"  x    \""),
+		},
+		{
+			give:             "x=:",
+			desc:             "invalid filters",
```
abhinav: nit:

> invalid: empty matcher

plan/config_test.go
```diff
@@ -116,3 +133,156 @@ func TestParseBehavior(t *testing.T) {
...
+		},
+		{
+			give:             "x=a:b,,c:d",
+			desc:             "invalid filters",
+			wantFilters:      nil,
+			wantBehaviorName: "",
+			wantError:        fmt.Errorf("match \"\" in input \"x=a:b,,c:d\" is not of form 'key:value'"),
+		},
+		{
+			give:             "x",
```
abhinav: Please add a test case for empty string

plan/entities.go
```diff
@@ -55,11 +58,37 @@ func (a Axes) Index() map[string]Axis {
...
 
-// Behavior represents the test behavior that will be triggered by crossdock
+// Filter specifies criteria for skipping specific test cases of a behavior.
+// All test cases for a behavior where all parameter values match the AxisMatches will be skipped.
+type Filter struct {
+	AxisMatches []AxisMatch
+}
+
+// AxisMatch matches an axis name to a give value.
+type AxisMatch struct {
```
abhinav: nit: Can we rename this struct to AxisMatcher

plan/entities.go
```diff
@@ -55,11 +58,37 @@ func (a Axes) Index() map[string]Axis {
 	return axes
 }
 
-// Behavior represents the test behavior that will be triggered by crossdock
+// Filter specifies criteria for skipping specific test cases of a behavior.
+// All test cases for a behavior where all parameter values match the AxisMatches will be skipped.
+type Filter struct {
+	AxisMatches []AxisMatch
```
abhinav: Can we rename this attribute to Matchers?

(This doesn't have to be manual. Your editor should take care of it for you.)

plan/entities.go
```diff
@@ -69,6 +98,22 @@ func (b Behaviors) Len() int           { return len(b) }
 func (b Behaviors) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
 func (b Behaviors) Less(i, j int) bool { return b[i].Name < b[j].Name }
 
+func (b Behaviors) validateAndApplyFilters(filtersByBehavior map[string][]Filter) error {
```
abhinav: call this `attachFilters` because that's really what it's doing. It's attaching the filters to the behaviors.

The filters are 'applied' later when we decide whether to skip a test case.

plan/entities.go
```diff
@@ -69,6 +98,22 @@ func (b Behaviors) Len() int           { return len(b) }
...
+	for i, behavior := range b {
+		filters := filtersByBehavior[behavior.Name]
+		for _, filter := range filters {
```
abhinav: Nit: `if len(filters) == 0 { continue }` to avoid changing the current behavior if it doesn't have filters

plan/entities.go
```diff
@@ -69,6 +98,22 @@ func (b Behaviors) Len() int           { return len(b) }
...
+			for _, axisToMatch := range filter.AxisMatches {
+				if !behavior.HasAxis(axisToMatch.Name) {
+					return fmt.Errorf("%v is not defined in axis for %v", axisToMatch, behavior.Name)
```
abhinav: change error message to:

> %q is not a parameter for behavior %q

plan/new.go
```diff
@@ -52,6 +57,27 @@ func buildTestCases(plan *Plan) []TestCase {
...
 				Arguments: testArgs,
+				Skip:      false,
+			}
+
+			for _, filter := range behavior.Filters {
+				// Each filter should be a complete match.
+				matchCnt := 0
+				for _, match := range filter.AxisMatches {
+					if (match.Name == behavior.ClientAxis && match.Value == client) ||
+						testArgs[match.Name] == match.Value {
```
abhinav: Most of this logic could be on the filter so that this just becomes,

    if filter.Matches(testArgs) {
      t.Skip = true
      ..
      continue
    }

You can even have an `AxisMatcher.Matches` function so that filter matcher just becomes,

    for _, m := range matchers {
      if !m.Matches(args) {
        return false
      }
    }
    return true

Also, note that testArgs already contains the ClientAxis and the behavior so you don't have to explicitly check for that.

plan/config_test.go
```diff
@@ -143,131 +143,115 @@ func TestParseSkipBehavior(t *testing.T) {
...
 		},
 		{
-			give:             "x=",
-			desc:             "invalid filters",
-			wantFilters:      nil,
-			wantBehaviorName: "",
-			wantError:        fmt.Errorf("match \"\" in input \"x=\" is not of form 'key:value'"),
+			desc:      "invalid: empty matcher",
+			give:      "x=:",
+			wantError: fmt.Errorf("invalid matcher \":\": axis name and value are required"),
```
abhinav: If you need to use quotes inside strings, use the ` form

    `invalid matcher ":": axis name and value are required`

This is the same as what you wrote except it's quite a lot more readable.

plan/entities.go
```diff
@@ -55,11 +58,55 @@ func (a Axes) Index() map[string]Axis {
...
+	Name  string
+	Value string
+}
+
+// Matches returns true if the given TestClientArgs match this AxisMatcher.
+func (a AxisMatcher) Matches(args TestClientArgs) bool {
+	if args[a.Name] == a.Value {
+		return true
+	}
+	return false
```
abhinav: Compress this whole thing to,

    return args[a.Name] == a.Value

plan/new.go
```diff
@@ -52,6 +57,20 @@ func buildTestCases(plan *Plan) []TestCase {
 				Plan:      plan,
 				Client:    client,
 				Arguments: testArgs,
+				Skip:      false,
+			}
+
+			for _, filter := range behavior.Filters {
+				// Each filter should be a complete match.
```
abhinav: obsolete comment

plan/new.go
```diff
@@ -52,6 +57,20 @@ func buildTestCases(plan *Plan) []TestCase {
...
+				if filter.Matches(testArgs) {
+					t.Skip = true
+					var formattedMatches []string
+					for _, match := range filter.Matchers {
+						formattedMatches = append(formattedMatches, fmt.Sprintf("%s:%s", match.Name, match.Value))
+					}
```
abhinav: This could be greatly simplified by a `String()` function on `AxisMatcher` and another on `Filter`.

abhinav (approved):
approved with nits

please fix before merging

PR #84, opened 2017-04-12
fix the build

vivekrsharma:
Got rid of unnecessary dependency on orchestra-io that caused build to fail.


kriskowal (approved):

kriskowal:
@breerly do you recall why we have this dependency? It may have been aspirational.

abhinav (approved):

bombela (approved):

